PR #22, opened 2015-07-27
added a progress bar to the valid route checker

alan-stokes:
added a progress bar to the vqalid route checker and cleaned up imports.


neworderofjamie:
Yep, looks fine

PR #26, opened 2015-11-09
Pull for buffering output

sergiodavies:
New classes to support output buffering functionality


rowleya:
Not now needed as no changes

PR #27, opened 2015-11-10
Work flow multi run

alan-stokes:
brings in the pause and resume modes of the tool chain from felix work into the tool chain proper. Works in the brnach, apart from changing the application graph mid runs.


alan-stokes:
this pull request contains a mother load of changes relating to multi-run, app_stop, reset on run and likely others ive forgotton. But bloody hell works now

PR #35, opened 2016-01-26
Router compressor

alan-stokes:
allows bigger simulations to operate as router entries will be compressed. This also includes vertex_key_alloc so that keys are partitionable_vertex|partitioned_vertex|atom instead of malloc based.

should allow 100% microcortical column to operate correctly.


mossblaser:
Hey there, I see you've got a slightly modified version of the BitField library in this PR? From my diff-ing the version you have differs from the Rig's latest release as follows:

```diff
@@ -13,7 +13,7 @@ from collections import OrderedDict
 from math import log
 
 
-class BitField(object):
+class RigsBitField(object):
     """Defines a hierarchical bit field and the values of those fields.
 
     Conceptually, a bit field is a sequence of bits which are logically broken
@@ -347,6 +347,10 @@ class BitField(object):
 
         return mask
 
+    def get_field(self, field_identifer):
+        internal_mapping = self._select_by_field_or_tag(field=field_identifer)
+        return internal_mapping[field_identifer]
+
     def _select_by_field_or_tag(self, tag=None, field=None):
         """For internal use only. Returns an OrderedDict of {identifier: field}
         representing fields which match the supplied field/tag.
@@ -477,7 +481,12 @@ class BitField(object):
                 self.field_values == other.field_values)
 
     def __ne__(self, other):
-        return not (self == other)
+        """
+        comparison  method for comparing constrains
+        :param other: instance of BitField
+        :return:
+        """
+        return not self.__eq__(other)
 
     def __repr__(self):
         """Produce a human-readable representation of this bit field and its
```

Aside from the `__ne__` fix (which thanks to @alan-stokes is now fixed as of project-rig/rig#218 and released in Rig v0.13.1), the only change is adding a "get_field" method which returns an internal `_Field` object. From poking around a little it appears you're using this to allow you to make fields whose position isn't assigned until the last minute? If I've understood correctly this functionality (auto-assigning fields to empty spaces in keys) is already part of the BitField library (see [`assign_fields()`](http://rig.readthedocs.org/en/stable/bitfield_doctest.html#rig.bitfield.BitField.assign_fields)). If I've not understood it would be helpful to learn what feature you're after since it would be ideal if this could be added to the Rig BitField implementation. Hopefully you should be able to use Rig's BitField implementation rather than forking/maintaining your own version :).

By the way, I'd be very wary of modifying internal data structures in BitField since it maintains a number of invariant properties of its internal structures which may easily be accidentally violated. Violation of these invariants can result in subtle but catastrophic bugs in key generation.

alan-stokes:
Howdi old boy, 

just to keep this in track. After a wee discussion, its come to light that the modifcation is not changing any data strucutres in bitfield, but more looking for the location of where a bitfield has been placed (from glances looking at code that ive forgotton in main memory these days) and so Jonathon is looking into providing this as some sort of correct api call for rig. 

I concur about not maintaining 2 version, this branch is still more than a first dash effort, and is still awaiting review, at which points like this should be recitfied. Thank you for reminding me of this, as it was on the list of stuff to talk to you about, but forgot with everythign else going on.
Alan

mossblaser:
Update: Rig v0.13.2 has an updated BitField object with a [`get_location_and_length` method](http://rig.readthedocs.org/en/stable/bitfield_doctest.html#rig.bitfield.BitField.get_location_and_length) you can use (safely!).

Make sure to update your setup.py to require at least this version! :)

rowleya:
Changes have now been done in adjustment of routing path

PR #37, opened 2016-02-16
Spreader partitioner with workflow

alan-stokes:
- [x] implement a spreader partitioner that  spreads the allocation of neurons to the entire machine, instead of sqeezing as many into the fewest cores.
- [x] test with basic runs
- [ ] merge with new dsg regione stimate brnaches
- [ ] retest with basic runs
- [ ] review completed

Currently works in principle, dse issues with memory estimates occurs which makes it currently a hit or miss algorithm, but changes in upcoming branches will make these issues go away.


rowleya:
This appears to create issues with the partitioned graph and the data generated.  If I run va_benchmark.py with this partitioner, I get a watchdog on chip when running the code during the reading of the master pop table.  Looking at the output, it produces 25 atoms per core, so I set va_benchmark.py to use 25 atoms per core and used the old Partitioner.  This worked correctly, leading me to assume something goes wrong in the graph when the SpreaderPartitioner is in action.

alan-stokes:
well least you got it to go further. I assume this has now got your memory calcs corrections? If so, I might have a regander at this after GFE branch is completed.

alan-stokes:
benching till we can get someone back on this

dkfellows:
Ping: are we ever going to come back to this or can we kill the branch?

alan-stokes:
killing as the work this was trying to solve has been targetted by bitfields and spreader placer. this is still technially a plasuable idea, but it probabily needs rewriting and the cost models of all our tools still have issues at levels of 1 neuron per core. so killing

PR #44, opened 2016-04-08
started possion heuristics

alan-stokes:
supprots splitting atoms via limited resoruces of the core for packets.
is being generised for multicast packets (both payload and none payload) fixed_route packets, sdram, cpu, dtcm.

[x] started interface
[ ] updated models to express their requirements 
[ ] updated resource tracker
[ ] updated partitioner
[ ] tested with simple test to verify nothing else goes boom.
[ ] tested with column to verify that it works in the hard case.
[ ] awaiting review


alan-stokes:
I'm using this place as a position holder, as it seems the most approiate for everyone to be able to contirbute their opinions to.

Ive been looking at the poisson splitting problem for about 3 days (on and off) now, and what was orginally thinking it was a sound hypothesis. Unfortnately now i feel it has fallen apart under deeper scurintiy. So im adding my logic here for everyone to digest and rip apart if judged that im losing my marbals(always likely, given shiny things and magpies). So lets start:

I started by assuming that we would need to have an accurate cpu calculation for a poisson generator for figuring when to split the rate. So I started looking into the c code with earnest, to deduce a better cpu estimate, than 0. What came out of it was that dependent upon the type of rate it was defined as, it would either call the rng a few times , 1 or 2, in terms of a fast source, where the distrubtion is uniform) or would take 5.2 times longer (in mean) for a slow  source.

This orginally got me thinking that for us to have issues at higher rate's, the generating of random number generators must be tied to the rate, as the rest of the code is a collection of for loops, based off spikes to send, which would have been very surprising to be the issue location, as they just wrote to a queue and fired off to the fabric, and if it was fabric failing to eat them, we'd have seen transmission queue provenance data items (which to my knowledge, was not the case).

I had a chat with michael hopkins about this, this morning, so that i could get correct estimates of the cpu cycles for the rng's themselves (not being mathimatically hot on impls of rngs) with regard to the rate and so put them into the python code for deducing when a poisson needed to be split. What actually came to light was that the generators were not linked to rate at all (apart from a multiplication whch is done after the generation) for either source, and that these cycle costs were no where near enough to account for the slippage of the poisson, as discussed below.

Michael believes that the rng used by the fast rate's should take between 40 and 50 clock cycles to generate its random number and should need to run twice ish, and therefore the slow system should take about 260 clock cycles per generator (5.2 * 50). Given that each core would have 200000 clock cycles when running at 0.1 ms time steps with a time scale factor of 10, this would argue that there could be 384 generators per core (assuming no other processing) and therefore we could assume about 1/2 overall given our extra overheads (384).  

This (if true) would mean that splitting the rates over multiple cores is pointless, as the cpu isnt going to save anything, as the slow rate is actually slower than the fast rate code, and its nearly basically constant cost no matter the rate. 

This motivated myself and michael to do a set of experiments trying out different combinations of rates, and generators per core to try and find where the issue really lay. I basically had 1 pop of Poisson feeding into 1 neuron in a IfCurExp pop. Below is the table of results:

| rate in hz | time step in millisecond | time scale factor | wall clokc tick in millisecond | runtime | issue summary | exp minus lamda | same for all generators | isi_val|
100 | 0.1 | 10 | 1 | 500 | 1 | n/a | n/a | yep | 100
1000 | 0.1 | 10 | 1 | 500 | 1 | n/a | n/a | yep | 10
10000 | 0.1 | 10 | 1 | 500 | 1 | n/a | n/a | yep | 1
20000 | 0.1 | 10 | 1 | 500 | 1 | weight saturated on if cur exp 1040 times | 0.1353352832 | yep | n/a
100000 | 0.1 | 10 | 1 | 500 | 1 | weight saturated on if cur exp 35200 times | 4.54E-005 | yep | n/a
100 | 0.1 | 10 | 1 | 500 | 1000 | n/a | n/a | yep | 100
1000 | 0.1 | 10 | 1 | 500 | 1000 | n/a | n/a | yep | 10
10000 | 0.1 | 10 | 1 | 500 | 1000 | timer tick on posson out of sync 150 times | n/a | yep | 1
 |  |  |  |  |  | timer on if cur exp overloaded 2640 times |  |  | 
 |  |  |  |  |  | weight saturated on if cur exp 931 times |  |  | 
20000 | 0.1 | 10 | 1 | 500 | 1000 | weight saturated on if cur exp 1040 times |  | yep | n/a
100000 | 0.1 | 10 | 1 | 500 | 1000 | n/a |  |  | 
 |  |  |  |  |  |  |  |  | 
10000 | 0.1 | 10 | 1 | 500 | 100 | n/a | n/a | yes | 1
10000 | 0.1 | 10 | 1 | 500 | 500 | timer on possion fall out of sync 150 times | n/a | yes | 1
10000 | 0.1 | 10 | 1 | 500 | 250 | n/a | n/a | yep | 1
10000 | 0.1 | 10 | 1 | 500 | 300 | n/a | n/a | yep | 1
10000 | 0.1 | 10 | 1 | 500 | 400 | weight saturated on if cur exp 2 times | n/a | yep | 1
10000 | 0.1 | 10 | 1 | 500 | 450 | timer on if cur exp overloaded 3124 times |  |  | 
 |  |  |  |  |  | weight saturated on if cur exp 7 times |  |  | 

The conclusion here was even at 100khz, we cant get 1 generator to fall out of timer step (we get the recievier to saturate, but thats not a big worry, given the weight was 2 and we had it firing every timer tick or more at those points). 

This means that splitting the rate between more cores in cases like the microcortical column (where the rate is between 10 and 20 khz) would not solve any problems (my opinion). 

Therefore myself and michael decided to explore the worst case (the point just before cores shift from slow to fast rate) which is exactly on 10khz rate to see how many neurons are acutally capable of being put into the system before the possion fails to operate correctly. After exploring from 100, up to 500. we have deduced its somewhere between 475 and 500. At 500, there is some effect from back pressure from the if cur failing to keep up with the input, but meh. 

During the process, it came to light that spreading the load of the If curr over multiple atoms and cores didnt gain you much, due to our sub pop routing (which required a bit of explination to michael, given its not his area of expertise), as we still send all the spikes from the generator to all cores and we only save the current processing, but still need to extract, input buffer, dma, drop work. This is annoyingly the same fundermental issue dan mandat found when moving fromd estination to soruce routing. This is going to be our major headache it seems (and maybe needs pointing out with neon lights to new devleopers....... i digress). 

This all comes to feed my hypthoesis that splitting rate of a poisson over multiple cores is a shed load of complexiity for little and maybe no gain. It may even make things worse, given the effect on back pressure by making packets flying around from multiple avenues of the system. Therefore this fix, i beieve boils down to adding a proper calculation for CPU for the poisson and adding cpu calcs to the resource tracker of the partitioner (which were not currently done, and need to be adjusted given wall clock timer ticks, as the processors measurement is for a 1ms time step, which isnt always valid) which would then allow it to aprtition the poisson correctly and remove this issue entirely. It may make more issues for the reciever. But if this is the case, then its not going to eb solved by splitting the poissons up anymore. We'd need to split the neuron model, (im sure thats not feasible) to go to neruon routing and explore keys and compression allocation a lot more. 

luckily most of these bits needed in the new hypothsis are the stuff I started on for the splitting code anyhow, so its not all wasted work cycles. 

By placing this logic here, I'm hoping that others in the team can read though and sound check my logic, because I feel this may be inportant for trying to figure the machines behaviour and allow us to focus better on solutiosn if this logic is sound.

regards
Alan

rowleya:
The reasoning behind splitting the Poisson source was not (at least directly) to do with the CPU usage, but rather the network resources.  The idea is that SpiNNaker is designed to run 1000 neurons per core at 10Hz.  Thus, it is expected that each core should output a maximum of 10,000 Hz.  In the microcircuit model, we have at points cores outputting around 20,000 Hz per neuron, and up to 500 neurons per core, which is quite a bit higher than this (even with 1 neuron it is twice the design specification).  I think that this can be achieved with the CPU, but this would then output more spikes than expected.  

I am aware that all these spikes will be going to the same place, but I am ignoring this for the time being; we should be able to cope with this with enough of a timescale factor from my calculations, but when running with a timescale factor of 1000, this still wasn't apparently enough when it should have been.

So the idea is now to divide the Poisson resources by not just the CPU, DTCM and SDRAM usage, but also by the network resources.  This could be done through a refactor of PACMAN to allow this additional parameter, and by changing the "atom" in the Poisson source to be the neuron-Hz rather than just the neuron - thus each Poisson can be split into a number of neuron-Hz per core.  If you want to avoid the PACMAN refactor, you might be able to get away with pretending that 10,000 Hz of Poisson is 100% of the CPU and then letting it divide things up in that way, but otherwise I would avoid looking at the CPU usage as part of this work (it probably should be done at some point later though).

rowleya:
I should add that the issue with the microcortical column was dropped packets more than timer overruns.  When run with DC we get neither dropped packets or (any significant) timer overruns.  With Poisson sources, we get dropped packets.  My guess is that these might be caused by the interaction with the source and the source router rather than anything else (but I have no proof of this).

alan-stokes:
Im wondering if this stuff is still needed....................... the grand problem of doing too many things, im not sure if this stuff is relvent anymore or has been merged in via other paths............ any thoughts?

rowleya:
Yes, doing something to scale the Poissons correctly would be a useful thing to have, though not urgent as we have it.  With the one-to-one master pop table, it could be useful to scale these to match the post-population when a one-to-one connector is used.

PR #47, opened 2016-06-17
new router for avoiding dead locks.

alan-stokes:
router which is designed to avoid deadlocks.
- [x] build new machine for wrap around basic machines using 0,1,2 as outgoing, 3,4,5 as incoming.
- [ ] build new machine for none wrap around using a circular order.
- [ ] test in both dimensions
- [x] add xml stuff


dkfellows:
This branch is dangerously broken in its current state. It appears to be not accounting for dead links in its search for whether there are wraparounds, and the rewriting of the machine model in one of the cases is completely and obviously wrong.

PR #48, opened 2016-06-17
Fix interface

rowleya:
Tidy up of the PACMAN interface
 - [x] Wait for merge of #45
 - [x] Fix outgoing edge partition to include the source vertex
 - [x] Fix routing information to be easier to access
 - [x] Fix graphs to be a single package for both partitioned and partitionable and fix naming
 - [x] Add AbstractEdge and AbstractVertex common to both partitionable and partitioned graphs and move OutgoingEdgePartition to use this
 - [x] Enforce the partition naming, as this is required for searching routing information
 - [x] Add ```__slots__``` notation to fix the interface and speed things up
 - [x] Change outputs to a list rather than a dict to allow normal calling of functions (and ensure the XML parsed order is preserved as required by the XML spec)
 - [x] Add support for injection of types
 - [x] Use interfaces where possible to work out what vertices need (e.g. tags)
 - [x] Add option of using decorators instead of XML for algorithm descriptions


alan-stokes:
"Fix graphs to be a single package for both partitioned and partitionable and fix naming". 

what logic is behind this? and given that we treat them differently, I'm rather curious on how your going to achieve this, or what the gain is in the end.

rowleya:
Yes, I will be more explicit.  There are a lot of common objects between them, such as the OutgoingEdgePartition.  I would also like to make an AbstractEdge and an AbstractVertex as there are a lot of places that we do things that work on both (like OutgoingEdgePartition).  This would solidify the structures all round.

rowleya:
Interestingly, this hints at moving the graph mapper in there too...

alan-stokes:
"Add support for dict as a constructor argument for data structures where an equivalent exists in rig and add a property to get the dict back, and support lazy operation of those structures such that the dict is not converted until a property (other than the dict itself) is requested"

what? this needs more baby speak for me to understand if ive done it or now. lol

PR #56, opened 2016-09-27
fixes injecotr to accept args of None

alan-stokes:
little bug that stops injections of params with value of none. fixed in this 
- [x] fixed
- [x] tested with script that exhibited error.


rowleya:
Appears to have already been done elsewhere!

PR #58, opened 2016-09-29
Reduced calls to get application vertex resources

rowleya:


alan-stokes:
this is a sub pull request from the same chip as constraint. As that one has been merged in, this one has become redundant.

PR #60, opened 2016-10-14
Same chip as constraint

rowleya:
Adds support for specifying that a group of Machine vertices must be placed on the same chip.  Currently this is supported by the RadialPlacer and the OneToOnePlacer, as well as RigPlace.


alan-stokes:
i did a few little changes to comments, removing a redundant param, and a expansion of a few set creators, as they confused me a bit until i commented and expanded them, and finally did a few param renames to make it more explicit (as i was looking and getting confused)

PR #67, opened 2017-02-03
Doc fixes for sphinx

Christian-B:
Ran shpinx locally and found errors.

Fixed an directory with and empty __init__.py and no other *.py class
Fixed :return: None where a : was missing
Fixed: Added missing line before parameters
Fixed: Added space after param <variableName>:


Christian-B:
Tested on readthedocs
See: http://pacman.readthedocs.io/en/doc_fixes_for_sphinx/

PR #69, opened 2017-02-06
Lazy data use

rowleya:
Makes more lazy in data structure use to make better when using a large machine (but not using all of it)

- [x] Wait for merge of SpiNNakerManchester/SpiNNMachine#31


rowleya:
Note, this is failing on Travis due to Dependence on SpiNNMachine changes

PR #70, opened 2017-02-13
directly use sphinx.autodoc.main

Christian-B:
using sphinx.apidoc.main works


Christian-B:
tested on read the docs so self approving to get master to work ASAP

PR #71, opened 2017-02-14
Fix automated build and add PEP8 enforcement

dkfellows:
This gets the test suite back into working order when building on Travis, records the requirements correctly, and adds enforcement of PEP8 rules via `flake8`.

It's a big change, but should not substantively change the API


rowleya:
An update on this, looking at the classes changed, the callers of the following need to be checked against the changes to OrderedSet:

- [ ] pacman.model.graphs.impl.outgoing_edge_partition.edges
- [ ] pacman.model.graphs.impl.graph.vertices
- [ ] pacman.model.graphs.impl.graph.get_edges_starting_at_vertex(vertex)
- [ ] pacman.model.graphs.impl.graph.get_edges_ending_at_vertex(vertex)
- [ ] pacman.model.graphs.impl.graph.get_outgoing_edge_partitions_starting_at_vertex(vertex)
- [ ] pacman.model.graphs.impl.graph.get_outgoing_edge_partitions_with_traffic_type(vertex)

PR #77, opened 2017-03-14
Corrections to the executor

dkfellows:
Some small fixes


pacman/executor/injection_decorator.py
```diff
@@ -86,19 +86,20 @@ def inject_items(types):
     :param types: A dict of method argument name to type name to be injected
     """
     def wrap(wrapped_method):
+        exn_arg = None
+        method_args = inspect.getargspec(wrapped_method)
+        for type_arg in types:
+            if type_arg not in method_args.args:
+                # Can't raise exception until run time
```
rowleya: Is there a reason why we can't raise the exception here?

PR #83, opened 2017-03-28
Made the key report easier to read

rowleya:


rowleya:
Fair enough - it is hard to do much of a test here, but I guess something could be done.

The main difference is that vertices are ordered in the graph whereas partitions are not.  Additionally, the amount of repeated information has been reduced, since each edge will have the same keys as every other edge in the partition.  This was making the report very long!

rowleya:
Maybe @alan-stokes could comment...

alan-stokes:
just reading email now. will look at request

alan-stokes:
so there's a little mismatch which makes this a bit problematic.

the network spec talks about verts and edges. It does not talk about edge
partitions. Therefore if i'm using the reports for debugging, the key
report needs to have a link to what edges are in the partition, or the
network report needs to describe what partition the edge is in.

This all said, just adding a list of edge labels in the key report under
each partition would feel the more easier to use location, as then you've
got all the data in one place.

regards
Alan

rowleya:
I see your point, but you will tend to add edges to the partition by specifying the partition id when you add them (which you have to do - this is now in the API), which it does provide e.g. it might say 
   
    Vertex: Bob
        Partition: SPIKES Key: 0x12345678

The list of edges was *very* long for the microcortical column...

rowleya:
Note that the new report won't be active until the merge of SpiNNakerManchester/SpiNNFrontEndCommon#117

PR #95, opened 2017-06-12
Better debugging so we can find failures in integration tests

dkfellows:
This should give us better info from the place that most integration tests are currently failing. A stack trace would be nice…


rowleya:
No longer relevant as the issue was found and fixed

PR #98, opened 2017-06-14
fixes issuse with pre_allocated resources and n chips required

alan-stokes:
this is needed for the chip-power-monitor, but is basically a bug fix that will need to go in anyhow, and until tis in, master is overestimating the chips required. 

issue is that orignally sdram tracker was used to figure chips use,d which before pre alloc was sound logic. Alas now sdram can be taken from pre allocated resources and not acutally needed in the n_chips. 

This fix tracks the chips which have had cores allocated, and therefore supports the system better. 

in chip-power-monitor, without this, the partitioner would state we need the biggest spalloc machine regardless of cores acutally used for the simualtion.


alan-stokes:
done. annoying these flake8's are mainly pycharm adding them during block comments, its not visible to the naked eye. would be nice if a pycharmer can figure out how to stop it from doing this by default

pacman/utilities/utility_objs/resource_tracker.py
```diff
@@ -1288,6 +1299,11 @@ def unallocate_resources(self, chip_x, chip_y, processor_id, resources,
         self._sdram_tracker -= resources.sdram.get_value()
         self._core_tracker[(chip_x, chip_y)].add(processor_id)
 
+        # check if chip used needs updating
+        if (len(self._core_tracker[(chip_x, chip_y)]) ==
+                self._machine.get_chip_at(chip_x, chip_y).n_processors):
+            self._chips_used.remove((chip_x, chip_y))
```
rowleya: Not sure this check is quite right.  Presumably you only want to remove it from chips_used if no processors are now used?  A test for what happens in various situations is really needed here...

alan-stokes: fair cope, ill build the tests once ive got the energy report tied down. currently implemting the number bits, all the boiler plate is in place now. so shouldnt be long before itll only need donals input, so ill be able to write it then

alan-stokes: fixed with test

alan-stokes:
ready for merging me thinks

Christian-B:
test this with sPyNNaker8/p8_integration_tests/test_various/test_many_spikes.py

Traceback (most recent call last):
  File "/brenninc/spinnaker/SpiNNFrontEndCommon/spinn_front_end_common/interface/abstract_spinnaker_base.py", line 997, in _run_algorithms
    self._shutdown()
  File ".../SpiNNFrontEndCommon/spinn_front_end_common/interface/abstract_spinnaker_base.py", line 2076, in _shutdown
    self._txrx.stop_application(self._app_id)
  File ".../SpiNNMan/spinnman/transceiver.py", line 1958, in stop_application
    process.execute(AppStop(app_id))
  File ".../SpiNNMan/spinnman/messages/scp/impl/app_stop.py", line 34, in __init__
    argument_2=(5 << 28) | _get_data(app_id, Signal.STOP),
  File ".../SpiNNMan/spinnman/messages/scp/impl/app_stop.py", line 11, in _get_data
    data = (_APP_MASK << 8) | app_id
TypeError: unsupported operand type(s) for |: 'int' and 'NoneType'

PR #99, opened 2017-06-27
Shorten names

dkfellows:
This is about shortening the sometimes over-long full names of classes.


pacman/operations/placer_algorithms/one_to_one_placer.py
```diff
@@ -219,22 +218,14 @@ def _find_one_to_one_vertices(vertex, graph):
...
                     next_vertex.constraints)
-                conflict = False
-                if x is not None and pre_x is not None and x != pre_x:
-                    conflict = True
-                if y is not None and pre_y is not None and y != pre_y:
-                    conflict = True
                 edges = graph.get_edges_starting_at_vertex(next_vertex)
-
-                if utility_calls.is_single(edges) and not conflict:
+                if is_single(edges) and not _conflict(x, y, post_x, post_y):
```
rowleya: This is broken - there is no post_x or post_y.  The left side refers to pre_x and pre_y - could this be what you mean?

PR #100, opened 2017-06-29
Redo the internal data structures to use slotted objects

dkfellows:
This reworks the code to stop using dictionaries of dictionaries. The use of constant string keys was particularly ugly!

**NB:** Needs someone to check that the tricks used actually work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ build/
 dist/
 PACMAN.egg-info
 /SpiNNaker_PACMAN.egg-info/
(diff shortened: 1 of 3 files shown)


rowleya:
Note that I have added a test which fails now.  The test passes on master, implying something has changed...

rowleya:
This seems to now pass the test so I think it can be merged... any reason not to?

alan-stokes:
not from what i can see

PR #105, opened 2017-07-04
Schema update

dkfellows:
This is the _actual_ schema used by the XML algorithm description files.

(FEC and sPyNNaker can't be changed to use the schema until we have the updated schema available properly on github.)


dkfellows:
I've checked, and the raw version (e.g., https://raw.githubusercontent.com/SpiNNakerManchester/PACMAN/e6564755a3c5b19805e382d77c1fc018b0d469a1/pacman/operations/algorithms_metadata_schema.xsd) is able to be used for validation purposes by Eclipse.

I don't want to use that particular version, but rather to point it at the version on `master` once that will actually validate things correctly.

PR #106, opened 2017-07-05
Schema urls

dkfellows:
Further work on the schema to support the XML as found in the rest of the sPyNNaker stack

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,11 @@ cache: pip
 virtualenv:
   system_site_packages: true
 
+addons:
+  apt:
+    packages:
+      - libxml2
+
 before_install:
   - pip install --upgrade pip setuptools wheel
   - pip install --only-binary=numpy,scipy numpy scipy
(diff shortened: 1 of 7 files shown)


dkfellows:
See also:
* https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/161
* https://github.com/SpiNNakerManchester/sPyNNaker/pull/346

dkfellows:
The XML reading code in Python does something strange when faced with XML Namespaces, even where those namespaces have the empty local prefix. Unfortunately, the only way to enable schema validation is to use XML namespaces, and so this all interacts horribly. This fixes it in an ugly way, but until https://bugs.python.org/issue18304 is fixed, we can't fix this in a better way.

* XML Validation against a schema is good.
* Schema validation requires namespaces be used (really!) because of weirdnesses in the W3C rules.
* Namespaces make Python's ElementTree spec change its behaviour quite radically.
* The changes in the ET spec (i.e., in what `Element.tag` reports) make our parsing code blow up.

Argh!

PR #115, opened 2017-09-11
Rig random

sara-es:
Random placement algorithm adapted from Rig's rand.py placer. Tracking of placed chips is handled in PACMAN's resource manager instead of in sets.


sara-es:
2017-09-11 10:41:08 INFO: Time 0:00:00.010405 taken by RandomPlacer

2017-09-11 10:44:29 INFO: Time 0:00:00.046910 taken by RigPlace
Time excluding conversion: 0:00:00.004949

Proving isomorphism is tricky because it's supposedly "random" - however both algorithms use random.sample(), might be possible to set a seed?

rowleya:
I believe you can use random.seed.  Note though that even in this case, you might call the random generator differently in the two cases.... worth trying though

alan-stokes:
sara mentioned this to me directly, and given the calls not being seeded and her saying she spent all morning looking for a way to seed the random sample call and failing to find a way to do so. I'm happy enough to just accept it if it looks random. 

given hers doesn't do the strange sampling with replacement stuff, its going to be impossible to compare exact results now. 

If it creates a random set of chips, and works, (and given its just a random placer) I'm happy to give it to her.

pacman/operations/rigged_algorithms/random_placer.py
```diff
@@ -0,0 +1,97 @@
...
+from pacman.utilities.algorithm_utilities import placer_algorithm_utilities
+from pacman.model.placements import Placement, Placements
+from pacman.model.constraints.placer_constraints import SameChipAsConstraint
+
+from spinn_utilities.progress_bar import ProgressBar
+
+import random as default_random
+
+
+class RandomPlacer(object):
```
alan-stokes: class doc

pacman/operations/rigged_algorithms/random_placer.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __call__(self, machine_graph, machine):
+
+        # check that the algorithm can handle the constraints
+        ResourceTracker.check_constraints(machine_graph.vertices)
+
+        placements_copy = Placements()
```
alan-stokes: why placements copy? there's no copy here. its not even able to make a copy, its a placer!

pacman/operations/rigged_algorithms/random_placer.py
```diff
@@ -0,0 +1,97 @@
...
+        # Iterate over vertices and generate placements
+        progress = ProgressBar(machine_graph.n_vertices,
+                               "Placing graph vertices")
+        resource_tracker = ResourceTracker(machine,
+                                           self._generate_random_chips(
+                                               machine))
+        vertices_on_same_chip = \
+            placer_algorithm_utilities.get_same_chip_vertex_groups(
+                    machine_graph.vertices)
+        all_vertices_placed = set()
```
alan-stokes: placed_vertices? over all_vertices_placed. its not got all to begin with, so ambigious

pacman/operations/rigged_algorithms/random_placer.py
```diff
@@ -0,0 +1,97 @@
...
+
+        # check that the algorithm can handle the constraints
+        ResourceTracker.check_constraints(machine_graph.vertices)
+
+        placements_copy = Placements()
+        vertices = \
+            placer_algorithm_utilities.sort_vertices_by_known_constraints(
+                machine_graph.vertices)
+
+        # Iterate over vertices and generate placements
```
alan-stokes: this comment needs to go 9 lines down. this should be something like "setting up data holders"

pacman/operations/rigged_algorithms/random_placer.py
```diff
@@ -0,0 +1,97 @@
...
+                vertices_placed = self._place_vertex(
+                    vertex, resource_tracker, machine,
+                    placements_copy,
+                    # placements,
+                    vertices_on_same_chip)
+                all_vertices_placed.update(vertices_placed)
+        return placements_copy
+        # return placements
+
+    def _check_constraints(
```
alan-stokes: static method. @staticmethod and remove self from list of params.

alan-stokes: also, who calls this? it isn't seeable from here.

pacman/operations/rigged_algorithms/random_placer.py
```diff
@@ -0,0 +1,97 @@
...
+            self, vertices, additional_placement_constraints=None):
+        placement_constraints = {SameChipAsConstraint}
+        if additional_placement_constraints is not None:
+            placement_constraints.update(additional_placement_constraints)
+        ResourceTracker.check_constraints(
+            vertices, additional_placement_constraints=placement_constraints)
+
+    def _generate_random_chips(self, machine, random=default_random):
```
alan-stokes: static method. @staticmethod above def and remove self from params

pacman/operations/rigged_algorithms/random_placer.py
```diff
@@ -0,0 +1,97 @@
+from pacman.utilities.utility_objs import ResourceTracker
+from pacman.utilities.algorithm_utilities import placer_algorithm_utilities
+from pacman.model.placements import Placement, Placements
+from pacman.model.constraints.placer_constraints import SameChipAsConstraint
+
+from spinn_utilities.progress_bar import ProgressBar
+
+import random as default_random
```
alan-stokes: why rename this? why not rename your variable so that its more readable and keep random as a known import

pacman/operations/rigged_algorithms/random_placer.py
```diff
@@ -0,0 +1,97 @@
...
+        :param machine: A SpiNNaker machine object.
+        :type machine: :py:class:`SpiNNMachine.spinn_machine.machine.Machine`
+        :param random: Python's random number generator
+        :type random: :py:class:`random.Random`
+        :return x, y coordinates of chips for placement
+        :rtype (int, int)
+        """
+
+        chips = set()
+        random.seed(2)
```
alan-stokes: this would become random_generator

pacman/operations/rigged_algorithms/random_placer.py
```diff
@@ -0,0 +1,97 @@
...
+        for x in range(0, machine.max_chip_x):
+            for y in range(0, machine.max_chip_y):
+                chips.add((x, y))
+
+        for _ in chips:
+            randomized_chips = random.sample(chips, 1)[0]
+            if machine.is_chip_at(randomized_chips[0], randomized_chips[1]):
+                yield randomized_chips
+
+    def _place_vertex(self, vertex, resource_tracker, machine,
```
alan-stokes: python doc

pacman/operations/rigged_algorithms/random_placer.py
```diff
@@ -0,0 +1,97 @@
...
+                      placements, location):
+
+        vertices = location[vertex]
+        # random x and y value within the maximum of the machine
+        chips = self._generate_random_chips(machine)
```
alan-stokes: this feels like repeated work. why cant you create a generator once, and keep using it? esp given your generator has the potential to repeat the same chips over and over again. Its almost like you want a while true generator. lol.

PR #122, opened 2017-09-28
Compression report

Christian-B:
Added a report which compares the compressed and uncompressed routing tables.

It check that every possible route in the uncompressed one is correctly covered in the compressed one.

### DIFF
--- a/pacman/operations/algorithm_reports/routing_compression_checker_report.py
+++ b/pacman/operations/algorithm_reports/routing_compression_checker_report.py
@@ -53,7 +53,7 @@ def codify(route, length=32):
 def codify_table(table, length=32):
     code_dict = OrderedDict()
     for route in table.multicast_routing_entries:
-        code_dict[codify(route)]= route
+        code_dict[codify(route)] = route
     return code_dict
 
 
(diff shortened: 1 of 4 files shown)


Christian-B:
Requires https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/205 to work but will not break anything if merged earlier.

pacman/operations/algorithm_reports/routing_compression_checker_report.py
```diff
@@ -0,0 +1,161 @@
...
+from pacman.operations.algorithm_reports import reports
+
+logger = logging.getLogger(__name__)
+
+
+WILDCARD = "*"
+LINE_FORMAT = "0x{:08X} 0x{:08X} 0x{:08X} {: <7s} {}\n"
+
+
+def codify(route, length=32):
```
rowleya: Does this function just return the string value of key & mask?

Christian-B: Instead of working with continuously with lots of keys, masks and binary operations what this does is convert the key and mask into a single string with 1, 0 or *.

Such that is only a 1 is accepted add a 1, if only 0 add a zero and use * if both are accepted.

This just speeds up comparison between two Strings instead of comparing two sets of keys and masks.

Christian-B: Or at least made is so I could understand what was going on!

PR #130, opened 2017-11-07
Improved Exception if clash between constraint and not idle

Christian-B:


pacman/utilities/utility_objs/resource_tracker.py
```diff
@@ -1261,8 +1262,13 @@ def get_maximum_resources_available(self, chips=None, processor_id=None,
                 best_processor_id = self._best_core_available(chip, key,
                                                               processor_id)
                 processor = chip.get_processor_with_id(best_processor_id)
-                max_dtcm_available = processor.dtcm_available
-                max_cpu_available = processor.cpu_cycles_available
+                try:
+                    max_dtcm_available = processor.dtcm_available
+                    max_cpu_available = processor.cpu_cycles_available
+                except AttributeError:
```
rowleya: That's very pythonic... would it be better to do the processor is None check afterwards?

pacman/utilities/utility_objs/resource_tracker.py
```diff
@@ -1261,8 +1262,13 @@ def get_maximum_resources_available(self, chips=None, processor_id=None,
...
+                    if processor is None:
+                        raise PacmanProcessorNotAvailableError(
+                            chip_x, chip_y, best_processor_id)
```
rowleya: After talking, this just needs a re-raise to ensure the original exception is raised if it isn't caused by None

pacman/utilities/utility_objs/resource_tracker.py
Christian-B: I this case NO. In all normally conditions None is not expected.
Therefor the code reads correctly in this should not happen but in case of emergency do....

I had wondered why chip.get_processor_with_id actually return None but as it specifically does assumed there was a valid use case so did not change that.

pacman/utilities/utility_objs/resource_tracker.py
Christian-B: How embarrassing!

I had assume I had done that as it is required.

Fixed.

pacman/utilities/utility_objs/resource_tracker.py
```diff
@@ -1310,8 +1311,14 @@ def get_maximum_resources_available(self, chips=None, processor_id=None,
...
-                max_dtcm_available = processor.dtcm_available
-                max_cpu_available = processor.cpu_cycles_available
+                try:
+                    max_dtcm_available = processor.dtcm_available
+                    max_cpu_available = processor.cpu_cycles_available
+                except AttributeError as ae:
+                    if processor is None:
+                        raise PacmanProcessorNotAvailableError(
+                            chip_x, chip_y, best_processor_id)
+                    raise ae
```
rowleya: Note there is a python trick: if you just call raise without arguments, it re-raises the existing error, which means it keeps its stack trace.

Christian-B: ok

PR #135, opened 2017-11-22
Add Tokens

rowleya:
This adds support for tokens to the workflow system.  Algorithms can now optionally generate a complete or partial token, and require a complete or partial token to have been generated (though it would be unusual to require a partial token).  The tokens are completely metadata-based; the algorithm code itself doesn't receive or return anything for these tokens to work.  This means that the workflow can be calculated upfront, without having to call any algorithm code.

This supports such things as having an algorithm that requires all possible parts of a process to have completed, but doesn't need to know how many such parts are available.  This is done through by looking through the algorithms for partial tokens generated in advance (including optional ones).  The token is then only regarded as complete when all these algorithms have been run.

It is also possible to override an algorithm using the tokens; simply put the algorithm that can be overridden in the optional algorithms and then the one that overrides it as required.  It will use the required algorithm first, and if it generates the same token, the optional one won't be run (unless it also generates another required unique token of course).  Additionally, an algorithm can generate more than one part and override more than one algorithm (which might still not be all algorithms).


pacman/executor/pacman_algorithm_executor.py
```diff
@@ -251,25 +290,28 @@ def _determine_algorithm_order(
         optionals_to_use = list(optional_algorithm_data)
         outputs_to_find = self._remove_outputs_which_are_inputs(
             required_outputs, inputs)
+        tokens_to_find = self._remove_complete_tokens(
```
alan-stokes: this is obsoleted, could be pushed into the building logic isntead of a seperate function thats only called once

pacman/executor/token_states.py
```diff
@@ -0,0 +1,87 @@
...
+
+    __slots = [
+
+        # The parts of the token that have not yet completed
+        "_incomplete_parts"
+    ]
+
+    def __init__(self):
+        self._incomplete_parts = set()
+        self._complete_parts = set()
```
alan-stokes: missing in __slots__

pacman/executor/pacman_algorithm_executor.py
rowleya: By that logic so could _remove_outputs_which_are_inputs could also be removed, but I think you will find that you added that one!

pacman/executor/token_states.py
rowleya: Fixed (and slots are now slots)

dkfellows:
The usual way of dealing with the “part” stuff is to put in a barrier operation that depends on all the other tokens and produces a single token when it is done. Such barriers are trivial algorithm blocks to define as they effectively do nothing.

rowleya:
OK, so this is avoiding the need for a barrier function, as that requires extra algorithms which have to be changed when you add an algorithm (or even worse, when the user adds an algorithm).  Users have to add 2 algorithms for the price of 1, which was part of what we are trying to avoid.  

This implementation avoids this by simply asking the algorithms if they are part of the process (as indicated by their token); if they are, they are just added to the things that need doing before that process is complete.

rowleya:
I guess this is the same as saying the barrier function you mention is now embedded by the workflow manager (this is the token_states object), and it is programmable based on the tokens that the algorithms generate and require.

alan-stokes:
having been scratching this a bit, it made a shed load more sense when i looked here:
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/blob/add_memory_file_interface/spinn_front_end_common/interface/interface_functions/front_end_common_interface_functions.xml

annoyingly, i really like it, having thought about it. the barrier is explicit in the xml, but means its isolated by each part from the others. so you only modify your algorithm to add anew part of the main token and bangs your uncle, its barrierised. No changing of the user of the token, or the other generators. 

i have a few issues with how its represneted in the xml, which rowley is going to fix tomorrow. but apart from that, im actualyl likeing the idea. that said, im going abck to the pacman code to see, now having understood it, how it works internally

rowleya:
Changes have all been done here now.  Additional work includes that the executor now will run optional algorithms that generate an optional input or token of a required algorithm before the required algorithm.  Optional algorithms are still only run if their outputs are novel.

dkfellows:
I quite like that tokens don't need to be passed as arguments or explicitly returned; since their values were never used, that's not a bad change.

pacman/operations/algorithms_metadata_schema.xsd
```diff
@@ -1,74 +1,93 @@
...
+        <xs:sequence maxOccurs="unbounded">
+            <xs:choice>
+                <xs:element name="param_name" type="xs:token"/>
+                <xs:element name="all_of" type="ParamSet" />
+                <xs:element name="one_of" type="ParamSet" />
+            </xs:choice>
+        </xs:sequence>
+    </xs:complexType>
+    <xs:complexType name="ResultList">
+        <xs:sequence maxOccurs="unbounded" >
```
dkfellows: Requires `minOccurs="0"` to support existing schemas.

rowleya:
Moved to #137

PR #138, opened 2017-12-06
Share key

alan-stokes:
supports sharing keys between partitions. Currently noone uses it. But it is required functionality for both NENGO into GFE and data speed up if we decide to go to MC paths instead of fixed route paths.


rowleya:
Failing in travis...

pacman/utilities/algorithm_utilities/routing_info_allocator_utilities.py
```diff
@@ -26,39 +32,542 @@ def get_edge_groups(machine_graph, traffic_type):
...
-    fixed_mask_groups = set()
-    fixed_field_groups = set()
-    flexi_field_groups = set()
-    continuous_groups = set()
+    fixed_key_groups = list()
+    shared_key_groups = list()
+    fixed_mask_groups = list()
+    fixed_field_groups = list()
+    flexi_field_groups = list()
+    continuous_groups = list()
```
dkfellows: I'm concerned about using a list instead of a set. Perhaps an OrderedSet would be better?

alan-stokes: 6 of one, 6 of an other. not really sure.

pacman/utilities/algorithm_utilities/routing_info_allocator_utilities.py
```diff
@@ -26,39 +32,542 @@ def get_edge_groups(machine_graph, traffic_type):
...
+    :rtype: None
+    """
+
+    # get types of constraints from this partition
+    (is_continuous, is_fixed_mask, is_fixed_key, is_flexi_field,
+     is_fixed_field, is_shared_key) = _check_types_of_constraints(
+        partition.constraints)
+
+    # this partition already in a share key somewhere, merge in
+    if partition in partition_mappings:
```
rowleya: I don't *think* the following code is wrong, but it is hard to follow...  A possible suggestion to make it easier is to realise that shared key is the only constraint that can group partitions.  Thus it might be easier to first group partitions and then define which of the other constraints apply to the group.

pacman/utilities/algorithm_utilities/routing_info_allocator_utilities.py
```diff
@@ -26,39 +32,542 @@ def get_edge_groups(machine_graph, traffic_type):
...
+            # remove mapping from wrong set, add to right set
+            group_currently_resides_in.remove(tracked_set)
+
+            # set tracked set to be new tracked set
+            tracked_set = new_tracked_set
+            real_group.append(tracked_set)
+
+            # update partition mapping accordingly
+            for partition in tracked_set:
+                partition_mappings[partition] = (tracked_set, real_group)
```
rowleya: This line is an example of the confusion here in - real_group contains tracked_set, but both are tracked here.  Is there a reason for this?

pacman/utilities/algorithm_utilities/routing_info_allocator_utilities.py
```diff
@@ -26,39 +32,542 @@ def get_edge_groups(machine_graph, traffic_type):
...
+
+        # if share key, locate and find partitions to merge
+        if is_shared_key:
+            # get share key constraint that must be there
+            other_partitions = \
+                utility_calls.locate_first_constraint_of_type(
+                    partition.constraints, ShareKeyConstraint).other_partitions
+
+            # search for merging groups
+            tracked_set, real_group = _search_for_merger(
```
rowleya: Again something that would be avoided by first grouping the partitions by shared key constraint

pacman/model/graphs/impl/outgoing_edge_partition.py
```diff
@@ -110,10 +110,19 @@ def traffic_weight(self):
         return self._traffic_weight
 
     def __repr__(self):
+        edges = ""
+        for edge in self._edges:
+            if edge.label is not None:
+                edges += edge.label + ","
+            else:
+                edges += str(edge) + ","
```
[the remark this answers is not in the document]
alan-stokes: was way way too much info when the edge str poured in, this is mainly a helpful reduction is what pycharm spits out allowing easier debugging. your welcome to remove / change accordingly

pacman/utilities/algorithm_utilities/routing_info_allocator_utilities.py
alan-stokes: it may be the only constraint that can add to a group, but by adding, you get the possibility of a cascade search. By handling each one separately, you (i think) reduce the amount of cycles, and allow you to consider just the addition of the constraints as they come. It also makes it faster to diagnose dodgy setups. But as i got myself scared of this code, simplification without performance reductions im happy for you to try. I'm still dripping from the wounds this gave me

pacman/utilities/algorithm_utilities/routing_info_allocator_utilities.py
alan-stokes: real group may not be the same tracked set. if your constraints are more than just share key, and you already reside in a group from previous share key links.

pacman/utilities/algorithm_utilities/routing_info_allocator_utilities.py
alan-stokes: but that grouping is a massive search, what if the next partition in the list has a seperate share key set which includes more and then repeat over and over. And now add if each of those partitions also include other constraints, which make it no longer a share key on tis own but say a fixed key share key set.

alan-stokes:
just to cover, im happy for someone to take another bash at this if you want. its scary as is

pacman/utilities/algorithm_utilities/routing_info_allocator_utilities.py
rowleya: You must already be doing this though to get the grouping - I will have a look and see what can be done...

pacman/utilities/algorithm_utilities/routing_info_allocator_utilities.py
rowleya: I will have a look...

pacman/utilities/algorithm_utilities/routing_info_allocator_utilities.py
alan-stokes: i do it piece meal, handling each partition as it comes. allowing us 1 pass.

pacman/utilities/algorithm_utilities/routing_info_allocator_utilities.py
rowleya: Having looked at this, they are now different things; in my changed version it is probably clearer that they are never added to twice in any case

pacman/model/graphs/impl/outgoing_edge_partition.py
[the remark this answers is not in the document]
rowleya: Don't care too much, just seemed odd - but OK :)

rowleya:
OK, I have done a commit that reworks this.  Comments welcome...

PR #139, opened 2017-12-06
fix harshness that fails spalloc machines

alan-stokes:
this removes the harshness of looking for optional inputs first. it results in spalloc machines failing as they run the chip allocater twice right after each other. This is ebcuase the pre alloc resources are designed to be ran in any order, and therefore havea  optional input which cripples the logic. 

by removing this restriction everything flows again. 

currently without this fix, master fails to run on spalloc machines due to being allocated a virutal board for doign real mappings on

### DIFF
--- a/pacman/executor/pacman_algorithm_executor.py
+++ b/pacman/executor/pacman_algorithm_executor.py
@@ -331,16 +331,11 @@ class PACMANAlgorithmExecutor(object):
             #  3. require optional inputs)
             order = [
 
-                # Check required algorithms forcing optional inputs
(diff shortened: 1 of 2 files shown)


rowleya:
Not only does this fail, but it isn't necessary.  It is simply possible to provide the appropriate optional token at the start; see SpiNNakerManchester/SpiNNFrontEndCommon#245.

A possible better solution would be to allow an algorithm to run if its optional requirements are not available but the algorithm itself provides that optional input as an output.  Another way of looking at this is that the generated output is not considered a generated output if the algorithm itself has the output as an optional input.

PR #142, opened 2017-12-11
Lint

dkfellows:
This is the first stage of linting: the simple bits.


pacman/executor/algorithm_classes/abstract_python_algorithm.py
```diff
@@ -50,13 +51,13 @@ def call(self, inputs):
             results = (results,)
 
         # If there are no results and there are not meant to be, return
-        if results is None and len(self._outputs) == 0:
+        if results is None and not self._outputs:
```
alan-stokes: how is that the same test?

pacman/executor/algorithm_classes/external_algorithm.py
```diff
@@ -20,6 +20,7 @@ def __init__(
             self, algorithm_id, required_inputs, optional_inputs, outputs,
             required_input_tokens, optional_input_tokens,
             generated_output_tokens, command_line_arguments):
+        # pylint: disable=too-many-arguments
```
alan-stokes: seriously? if theres a limit to arguments, then maybe we should either consider reducing the length, or asking why pylint has such a limit. Given having that all over the place i feel is almsot as bad as having the warnings

pacman/executor/algorithm_classes/python_class_algorithm.py
```diff
@@ -61,10 +60,10 @@ def call_python(self, inputs):
             method = "__call__"
             if self._python_method is not None:
                 method = self._python_method
-            exc_type, exc_value, exc_trace = sys.exc_info()
-            logger.error("Error when calling {}.{}.{} with inputs {}".format(
-                self._python_module, self._python_class, method, inputs))
-            raise exc_type, exc_value, exc_trace
+            logger.error("Error when calling %s.%s.%s with inputs %s",
```
alan-stokes: this goes against python recommendations of using .format. So why does pylint demand it?

pacman/executor/algorithm_classes/python_function_algorithm.py
```diff
@@ -44,10 +44,9 @@ def call_python(self, inputs):
         try:
             return function(**inputs)
         except Exception:
-            exc_type, exc_value, exc_trace = sys.exc_info()
-            logger.error("Error when calling {}.{} with inputs {}".format(
-                self._python_module, self._python_function, inputs))
-            raise exc_type, exc_value, exc_trace
+            logger.error("Error when calling %s.%s with inputs %s",
```
alan-stokes: same issue about not using .format here

pacman/executor/algorithm_decorators/token.py
```diff
@@ -31,6 +31,7 @@ def __hash__(self):
         return str(self._name + ":" + str(self._part)).__hash__()
 
     def __eq__(self, other):
+        # pylint: disable=protected-access
```
alan-stokes: obviously the correct fix here is to have porperties for those privates, not just letting protected stuff be circumvented

pacman/executor/injection_decorator.py
```diff
@@ -18,6 +18,7 @@ def supports_injection(injectable_class):
     orig_init = injectable_class.__init__
 
     def new_init(self, *args, **kwargs):
+        # pylint: disable=protected-access
         orig_init(self, *args, **kwargs)
         for method in injectable_class.__dict__.itervalues():
```
alan-stokes: wow, this defintately should be a porperty, not a call to the private __dict__!

pacman/executor/pacman_algorithm_executor.py
```diff
@@ -667,39 +661,39 @@ def algorithm_timings(self):
     def _update_timings(self, timer, algorithm):
         time_taken = timer.take_sample()
         if self._print_timings:
-            logger.info("Time {} taken by {}".format(
-                str(time_taken), algorithm.algorithm_id))
+            logger.info("Time %s taken by %s",
```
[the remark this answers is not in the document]
rowleya: Note that if we want to keep the format strings, but not actually have them evaluated (which is expensive in time; even log.debug messages with .format are evaluated), there is a solution:
https://stackoverflow.com/questions/13131400/logging-variable-data-with-new-format-string/24683360#24683360

pacman/executor/algorithm_classes/abstract_python_algorithm.py
rowleya: This *is* the same test in Python...

pacman/executor/algorithm_classes/external_algorithm.py
rowleya: I don't think having too many arguments is something that can really be helped....

pacman/executor/algorithm_decorators/token.py
rowleya: This is accessing items that are part of the class, which has been type-checked, so it is fine here

pacman/executor/injection_decorator.py
rowleya: This is acceptable in this context - decorators sometimes have to do some dirty work!

rowleya:
A small test shows that this seems to work, so merging...

pacman/executor/algorithm_classes/external_algorithm.py
dkfellows: I treat “too many arguments” (and even more so “too many locals”) as a marker for where I need to see if there is an opportunity to refactor. Often it isn't practical, but I'd rather have the false positives as this is not in a tool that we're using for full enforcement.

pacman/executor/algorithm_decorators/token.py
dkfellows: Agreed with @rowleya; if this was Java or C#, we wouldn't have the complaints at all.

PR #143, opened 2017-12-12
Fix partitioning

rowleya:
Fixes partitioning when not using extra monitor support


alan-stokes:
so i can respect this pr, but during the review, i checked out the fake inputs and tokens (which i msut have checked before, but not done properly). why do we need fake ones? it feels like a bodge

rowleya:
To address the comment, the fake things are a way of allowing optional things to become required in the presence of an ability to generate them i.e. if I can generate something that is optional, I should make every attempt to do so before running an algorithm that takes that thing as an optional input.  The "fake" things are those things which are optional and *cannot* be generated by the workflow and thus will never be available.  By having the fake inputs and tokens you make the flow easier to follow.

PR #154, opened 2018-04-20
SDRAM edges

rowleya:
This adds support for SDRAM edges - edges which have an SDRAM traffic type, and thus indicate communication over SDRAM.  This is similar to the SameChipAsConstraint, but has the following advantage: during partitioning, SDRAM edges will be expanded up to full connectivity.  The application can then further filter edges which don't match the appropriate connectivity.

An example:
Start with application vertices:

A(2 atoms) -> SDRAM edge -> B(2 atoms) 

with max 1 atom per core for each, and SDRAM communications on a 1-1 basis.  Partitioner will create machine vertices and edges:

A1 -> SDRAM edge -> B1 (needed)
A1 -> SDRAM edge -> B2 (can be filtered)
A2 -> SDRAM edge -> B1 (can be filtered)
A2 -> SDRAM edge -> B2 (needed)

With a SameChipAsConstraint, there would be no such ability to filter.

Another example:
Start with application vertices:

A(2 atoms) -> SDRAM edge -> B(4 atoms)

which should be connected in the ratio of 1 atom of A to 2 atoms of B.  The partitioner will create:

A1 -> SDRAM edge -> B1,2 (needed)
A1 -> SDRAM edge -> B3,4 (can be filtered)
A2 -> SDRAM edge -> B1,2 (can be filtered)
A2 -> SDRAM edge -> B3,4 (needed)

Again a SameChipAsConstraint would offer no ability to filter the edges.  Similarly assuming that SameChipAsConstraint applies on an atom-by-atom basis would fail.  The SDRAM edge offers a more flexible solution,

### DIFF
--- a/pacman/model/graphs/common/edge_traffic_type.py
+++ b/pacman/model/graphs/common/edge_traffic_type.py
@@ -7,3 +7,4 @@ class EdgeTrafficType(Enum):
 
     MULTICAST = 1
     FIXED_ROUTE = 2
+    SDRAM = 3
--- a/pacman/operations/placer_algorithms/one_to_one_placer.py
+++ b/pacman/operations/placer_algorithms/one_to_one_placer.py
@@ -7,8 +7,8 @@ from pacman.exceptions import \
 from pacman.model.placements import Placement, Placements
 from pacman.operations.placer_algorithms import RadialPlacer
 from pacman.utilities.utility_objs import ResourceTracker
-from pacman.utilities.algorithm_utilities \
-    import placer_algorithm_utilities as placer_utils
+from pacman.utilities.algorithm_utilities.placer_algorithm_utilities \
+    import get_same_chip_vertex_groups, sort_vertices_by_known_constraints
 from pacman.model.constraints.placer_constraints\
     import SameChipAsConstraint
 from pacman.utilities.utility_calls import is_single
@@ -37,8 +37,7 @@ class OneToOnePlacer(RadialPlacer):
             additional_placement_constraints={SameChipAsConstraint})
 
         # Get which vertices must be placed on the same chip as another vertex
-        same_chip_vertex_groups = placer_utils.get_same_chip_vertex_groups(
-            machine_graph.vertices)
+        same_chip_vertex_groups = get_same_chip_vertex_groups(machine_graph)
         sorted_vertices = self._sort_vertices_for_one_to_one_connection(
             machine_graph, same_chip_vertex_groups)
 
@@ -120,8 +119,7 @@ class OneToOnePlacer(RadialPlacer):
         found_list = set()
 
         # order vertices based on constraint priority
-        vertices = placer_utils.sort_vertices_by_known_constraints(
-            machine_graph.vertices)
+        vertices = sort_vertices_by_known_constraints(machine_graph.vertices)
 
         for vertex in vertices:
             if vertex not in found_list:
--- a/pacman/operations/placer_algorithms/radial_placer.py
+++ b/pacman/operations/placer_algorithms/radial_placer.py
@@ -1,8 +1,8 @@
 # pacman imports
 from pacman.model.constraints.placer_constraints \
     import RadialPlacementFromChipConstraint, SameChipAsConstraint
-from pacman.utilities.algorithm_utilities \
-    import placer_algorithm_utilities as placer_utils
+from pacman.utilities.algorithm_utilities.placer_algorithm_utilities \
+    import sort_vertices_by_known_constraints, get_same_chip_vertex_groups
 from pacman.model.placements import Placement, Placements
 from pacman.utilities.utility_calls import locate_constraints_of_type
 from pacman.utilities.utility_objs import ResourceTracker
@@ -28,16 +28,14 @@ class RadialPlacer(object):
         self._check_constraints(machine_graph.vertices)
 
         placements = Placements()
-        vertices = placer_utils.sort_vertices_by_known_constraints(
-            machine_graph.vertices)
+        vertices = sort_vertices_by_known_constraints(machine_graph.vertices)
 
         # Iterate over vertices and generate placements
         progress = ProgressBar(
             machine_graph.n_vertices, "Placing graph vertices")
         resource_tracker = ResourceTracker(
             machine, self._generate_radial_chips(machine))
-        vertices_on_same_chip = placer_utils.get_same_chip_vertex_groups(
-            machine_graph.vertices)
+        vertices_on_same_chip = get_same_chip_vertex_groups(machine_graph)
         all_vertices_placed = set()
         for vertex in progress.over(vertices):
             if vertex not in all_vertices_placed:
--- a/pacman/operations/rigged_algorithms/hilbert_placer.py
+++ b/pacman/operations/rigged_algorithms/hilbert_placer.py
@@ -48,8 +48,7 @@ class HilbertPlacer(object):
             machine, self._generate_hilbert_chips(machine))
 
         # get vertices which must be placed on the same chip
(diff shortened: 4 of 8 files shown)


alan-stokes:
looks fine to me. knew it wouldn't be too complex to tie into the placer. 

Now, as a future theory, which is related to this on a off shoot. sdram edge may require sdram allocation, which means the dsg needs to handle it differently. i think. So there may be a need for a special edge type for this, as it may require sdram size parameter to be stored.

rowleya:
Adding SDRAM to the edge would mean that the placer would need to take this into account as well I guess.  As you say, there would then need to be something that allowed the allocation of the requested space and then something might need to write data to that space (application dependent).  I think I would leave this as future work for now...

alan-stokes:
so through i concur its a future work. my immediate question is how is Roberts stuff using the sdram link? As there must be some sdram shared between these cores somehow, and therefore  whatever that is, would need to change when we've set it up. so doing that now would seem prudent.

rowleya:
Currently Robert's stuff is using SameChipAsConstraint on MachineVertex instances.  He then just picks one of those MachineVertex instances to allocate the shared SDRAM space.  I am not absolutely sure how the communication of the shared space is done between the vertices, but I believe that either the address or the tag is sent using a multicast message.  In any case, this is currently handled in a custom way that works, and should continue to work with this change; the only thing this change really does is allow for ApplicationVertex instances with filterable same-chip-as constraints.

Speaking of this, an ApplicationVertex test might be a good idea....

alan-stokes:
a custom way is all fine and good when its a stand alone thing, but we're meant to be figuring out how to support this in general, at which point a custom way, may become the norm, and/or become inconsistent as times go by. thus thinking about it now and modifying as required is a better solution than "it works, lets not mess with it"

app vertex test with a pruner would be good.

rowleya:
I agree in principle, but this needs a bit of thought on how to make it work.  

On a MachineEdge it is simple enough - there can be a resources_required which returns the SDRAM to be used (along with other things as necessary).  The assumption here would be that *any* resources returned by a MachineEdge *must* be shared between source and target vertex - thus they must be on the same chip.  This can then be added in when placing the vertex.  The SDRAM traffic type is now a tag really to say that the edge doesn't need to be run through a router.

On an ApplicationEdge this is harder; we now have to do the partitioning on both the vertices of the edge, and on the edge itself at the same time I think.  That would then make sure that both fit on a single chip together.  That isn't trivial in itself (as some heuristic has to be used to decide how to break down the two sides potentially independently of each other).

alan-stokes:
i concur, thus why we're thinking. 

my personal thoughts was  that youd take all the app verts which have a sdram link and do these before any other vertex. you'd do same splitting (like we do with delays) for all the destinations of all the sdram links coming out of the vertex your splitting. The sdram cost of the links is each edges sdram cost as you go. once youve split part of it, unless there's a circular sdram link (which would be very dubious at best) then it doesn't matter where the next sets reside, same chip as the last split, or another chip. Just as long as theres enough cores/sdram on the considered chip for the next set.
if there is a recursive sdram link, then id argue, split them all as a normal set, and then if they take up more sdram than available (when sdram links are considered), then  it was never going to work, so boom it goes anyhow.

id say that before dsg, a phase of sdram alloc for sdram edges are done, and the memory address stored in the edge (this also means the sdram edge needs a chip placement stored within it). The associated verts of the edge can then during dsg store the address as init data, and bobs ya uncle, works as planned.  

what ya think?

rowleya:
For the partitioning, you don't want to enforce the same atoms on both sides of the edge, as this might not work.  This means you really have 2 sets of atoms to reduce at the same time, and each time test against the requirements.  We can calculate the requirements of vertices f and g, and edge h, with n_atoms a and b respectively, as f(a) + g(b) + h(a, b) and compare this to the space available s.  Thus ratio = (f(a) + g(b) + h(a, b)) / s.  We now need to reduce by approximately the ratio, and we can make the same assumption as currently that the space is proportional to the number of atoms (which may not actually be true, but this is trial and error).  Probably the easiest way to achieve something like this (not precise but it might do) is to pick the largest of a and b and reduce that by the ratio, then reduce the other by the ratio of the new value to the old value e.g. if a > b then a_new = a / ratio, b_new = b / (a / a_new).  This will maintain the ratio between the numbers.  Once you reach an acceptable place, you can then start increasing the numbers to get something closer to correct.

I guess the sdram edges will now need placements as well (we can add placement of edges to the placements object).  

There could also be a memory allocation phase.  I would suggest that this is the part left as future work.

alan-stokes:
that reads as  if your assuming 1 sdram edge. theres nothign stopping you having many sdram edges to many verts. so its a rather bigger problem than just considering 2 verts and one edge. 

thats why i considered them all being split equally, and even though i cant see a major issue with splitting equally. i do concur its a built in assumption. so assuming you expand your logic to be lsits of verts and lists of edges, im in for parittioning

having thought about putting edges into placements, i really wouldnt. thats going to make placements way more complicated, as a placement assuming x,y and P. a edge has no P. What we can do, is during sdram alloc for the sdram edge, is ask for the placement of source and/or destination vertex. that then keeps it all clean....... but the sdram edge will need the memory address. or you need to build a map of sdram edge to memory address and hand that down to all dsg's. which means either more api changes, or more injections to be done.......

alan-stokes:
unknown where this is right now?

rowleya:
Work in progress - I think everything is checked in so you are welcome to look at it :).

alan-stokes:
as in, work in progress, dont review / merge if happy. or work done, ready for review and merge if happy?

rowleya:
I think I was at the point where I was working out how to integrate the SDRAM edge with the partitioner in such a way that the vertices are grouped for partitioning, and the edge is given resource for that partitioning in some way.  So don't merge yet...

alan-stokes:
ill leave ya to it then. wont bother reviewing till its polished. ill go to that retina partitions solution now then.

alan-stokes:
"as have resources for edges." 

did we?! thoguth the nengo was going to have said resources on the sdram edges, and then the parttion handle the allocation. did we have different convos?

rowleya:
I think this sounds like a specific instance issue i.e. you can do that in your case, but this doesn't generalise.  The problem is that if the edge has resources, my only choice in the resource tracker is to assume each edge requires that resource.  I have no information that tells me that the resource itself is shared, hence it is easier to leave the resource to the source and/or target.

rowleya:
If you are happy with that solution (i.e. assuming that the resource is per-edge and not shared), it can be added of course...

alan-stokes:
whereas my issue is there is no sharing of said resource. it is per edge. But the entire block of those edges need to be allocated as one block of sdram. *sighs* always complicated.

alan-stokes:
"bit of a pain for partitioning, as you then need to partition the pre- and post-vertices together as previously mentioned. " you have meet this nengo stuff. partitioning an ensemble requires all its machine verts to be done at the same time, as its resources required are affected by the number of splits. So im dealing with that anyhow. 

and I thought we'd said that 1 edge woudl take the resoruces and rest would have zero for robs version. anyhow. if the edge isnt to have sdram. ill overload it and add one for the nengo version.

rowleya:
Even with just one edge having the resources, partitioning becomes a pain.  As you suggest, you can deal with it at the application level in any case.

If you are happy with these changes though, feel free to hit the go button on this pull request!

alan-stokes:
added to my todo list. right after lif partitioning sdram calc solved. As this is then to be tied into this. I've already lost the blasted region which has the locks too much code being reviewed.

alan-stokes:
reviewed. rubbering once ticks have reappeared (after merging master in)

PR #155, opened 2018-04-21
Min atoms per core

rowleya:
Add a constraint for min atoms per core.  This is useful in a scenario where the number of atoms is useful for reason other than simple divide-ability.

Note that I tried adding this to the BasicPartitioner too, but the way it works makes this quite difficult to make it work correctly!

### DIFF
--- a/pacman/model/constraints/partitioner_constraints/__init__.py
+++ b/pacman/model/constraints/partitioner_constraints/__init__.py
@@ -1,7 +1,9 @@
 from .abstract_partitioner_constraint import AbstractPartitionerConstraint
 from .max_vertex_atoms_constraint import MaxVertexAtomsConstraint
+from .min_vertex_atoms_constraint import MinVertexAtomsConstraint
 from .same_atoms_as_vertex_constraint import SameAtomsAsVertexConstraint
 
 __all__ = ["AbstractPartitionerConstraint",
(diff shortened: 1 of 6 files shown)


alan-stokes:
i think this is slightly flawed. please create a test that does the following.

create a vertex with 10 atoms, with a min atoms of say 5, create a machine which when the first time it tries to allocate, fails, then runs the ratio calc (this is where id need a calculator to do the maths), but you want the sdram requirements and max sdram to be such that the ratio results in the next partition being 4 atoms, but the 5 atoms system would actually work. i'm not sure sure if my theory is clear.

but i think the partitioner would not try to up scale to min atoms, and therefore would fail, when it can indeed be done.

rowleya:
OK, I have added a test and it works, but I'm not sure if this is quite what you are after.  The issue is that I can't get it to have a ratio that means that it will try 4 when 5 will work.  If 5 works, it tries 5 first seemingly.  I guess this might mean that the maths gets around the issue (as the ratio is based on the space available).  Happy to try something else if you want though...

rowleya:
OK, I managed to find the error case that fails when it shouldn't.  Having talked about it with Christian, we are wondering if a "FixedAtomsPerCore" constraint might be better given the use case.

rowleya:
Closing as Fixed atoms per core is preferred.  See #156.

PR #156, opened 2018-04-23
Fixed atoms per core

rowleya:
Added a FixedAtomsPerCore constraint.  This enforces the splitting of an ApplicationVertex into fixed sized chunks.  Max atoms per core is ignored except to check that it is valid.

This works well on vertices where the concept of an atom has more meaning than just the subdivisability of the vertex in partitioning.


alan-stokes:
this comes from roberts spinnaker ear. which to go through the partitioner as a pynn set, needs extra constraints due to its operational assumptions. its loks fine to me too.

PR #159, opened 2018-05-03
vertex are sorted by label so give a default label

Christian-B:


pacman/model/graphs/machine/machine_vertex.py
```diff
@@ -26,7 +26,10 @@ def __init__(self, label=None, constraints=None):
             * If one of the constraints is not valid
         """
         super(MachineVertex, self).__init__(constraints)
-        self._label = label
+        if label:
+            self._label = label
+        else:
+            self._label = str(type(self))
```
dkfellows: I'd slightly prefer this one-liner:
```
self._label = label if label else str(type(self))
```
as that fools the coverage checker into thinking it's covered.

PR #160, opened 2018-05-10
Better no alogithm

Christian-B:


pacman/executor/pacman_algorithm_executor.py
```diff
@@ -432,7 +431,8 @@ def _determine_algorithm_order(
                     "    Tokens complete: {}\n"
                     "    Fake tokens complete: {}\n"
                     "    Tokens to find: {}\n"
-                    "    Functions available: {}\n"
+                    "    Algorithms Left: {}\n"
```
dkfellows: It might be better to label this `Required algorithms remaining to be used:` or something like that.

Christian-B: Agreed todo

pacman/executor/pacman_algorithm_executor.py
```diff
@@ -391,27 +391,26 @@ def _determine_algorithm_order(
...
-                        algorithm_input_requirement_breakdown += \
-                            self._deduce_inputs_required_to_run(
-                                algorithm, input_types, token_states,
-                                fake_inputs, fake_tokens)
+                    algorithm_input_requirement_breakdown += \
+                        self._deduce_inputs_required_to_run(
+                            algorithm, input_types, token_states,
+                            fake_inputs, fake_tokens)
                 for algorithm in optionals_to_use:
-                    if algorithm.algorithm_id in algorithms_left_names:
```
alan-stokes: by removing this and the one above, i THINK (but to be honest, id need to run tests to verify) that you've opened up a risk of running the same algorithm multiple times. But it might be picked up later on due to not producing any new outputs.......but not sure.

rowleya: By this point, an error is about to be produced I think...

alan-stokes: having opened the entire file, yep ya right. 
nail in why i hate using githubs interface. lol. i dont see it in all context. 

on that note, happy to merge

PR #163, opened 2018-05-22
ResourceTracker handling multiple pres for one chip

Christian-B:


Christian-B:
Required for https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/329 but does not break if merged first

PR #167, opened 2018-05-24
How to generate all masks, not just the one with the zeros at the bottom

dkfellows:
This uses cleverness, functools and itertools to let us try all possible masks.


dkfellows:
This does generate the value that we used to get first, so if that's OK then this all works as before…

rowleya:
This would be fine if it stopped generating at that point.  The issue is that if it doesn't find keys in the contiguous space, it will then go on to try non-contiguous keys.  The answer to this is probably to have a Boolean argument for contiguity, and then have the algorithm fall out if contiguous keys are requested.

PR #176, opened 2018-08-09
fixes screw up with resource tracker for virtual chips

alan-stokes:
this tracks the virtual chips separately to the real chips. It also makes the constant for n_cores_per_virtual_chips a actual constant, and not a magic number (who did that in the first place ;-) ) 

It should fix the issue arren was seeing. It also allows users to ask for the vc with the most usable cores.......if this is usful........im not sure......... but as it comes out of the changes. it might as well be put in.

### DIFF
--- a/pacman/utilities/utility_objs/resource_tracker.py
+++ b/pacman/utilities/utility_objs/resource_tracker.py
@@ -4,7 +4,7 @@ from pacman.model.constraints.placer_constraints\
     import ChipAndCoreConstraint, AbstractPlacerConstraint
 from pacman.model.resources import ResourceContainer, DTCMResource, \
     SDRAMResource, CPUCyclesPerTickResource
-from pacman.utilities import utility_calls
+from pacman.utilities import utility_calls, constants
 from pacman.exceptions import PacmanInvalidParameterException, \
     PacmanValueError, PacmanException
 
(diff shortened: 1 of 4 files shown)


pacman/utilities/utility_objs/resource_tracker.py
```diff
@@ -1211,17 +1229,29 @@ def _available_resources(self, usable_chips):
...
+        :return: the max cores available on the best real chip
+        :rtype: int
+        """
+        for n_cores_available, n_chips_with_n_cores in reversed(list(
+                enumerate(self._real_chips_with_n_cores_available))):
+            if n_chips_with_n_cores != 0:
+                return n_cores_available
 
-        :return: the max cores available on the best chip
+    def get_maximum_cores_available_on_a_virtual_chip(self):
```
rowleya: Most of this change is for this call - will you ever call this?  If not, the code just needs to keep the checks for virtual, but can miss out the actual count itself.

alan-stokes: as i mentioned in the commit message and pr:

It also allows users to ask for the vc with the most usable cores.......if this is usful........im not sure......... but as it comes out of the changes. it might as well be put in.

and given the code will only run on the vm chips, its not costing us much. I'm happy to take it out if your sure noone will ask. Just seemed a cheap bell (and yes i note what im saying to the standard software engineer)

rowleya: OK, happy enough then - it currently fails tests though...

alan-stokes: no idea what ya on about *cheeky smile* bloody f8

PR #177, opened 2018-08-22
Support for iptags no ports

alan-stokes:
this functionality allows iptags to not have ports too. not fully tested and have fear that its missing something. but it allows nengo feeds to in theory have multiple instances, if i decide to go down iptag level. Currently leaving it as sdp messages due to the limit of iptags (note we might want a more flexible way to allocate iptags space on the monitor core).

### DIFF
--- a/pacman/model/graphs/impl/graph.py
+++ b/pacman/model/graphs/impl/graph.py
@@ -174,7 +174,7 @@ class Graph(ConstrainedObject, AbstractGraph):
 
     @overrides(AbstractGraph.get_outgoing_partition_for_edge)
     def get_outgoing_partition_for_edge(self, edge):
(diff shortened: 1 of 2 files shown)


rowleya:
So it is already possible to have IP tags without ports - instead you add a "traffic type" to the tag resource, and all things with the same traffic type will then be given the same tag.  The vertex that requested the tag is then meant to look at the IP tags it has been assigned and make sure that the port is set.

What this seems to do is add an assignment of a port to those IP tags that don't have ports.  The issue here is that the port of the IP Tag actually has to mean something i.e. it should actually target an open port on the host machine.  I don't think this can be done automatically by the tools.

rowleya:
Closing as seems wrong.

PR #178, opened 2018-08-22
Partition map by edge

alan-stokes:
allows the graph to track what edge is assocated with each partition. used by nengo extensively.

### DIFF
--- a/pacman/model/graphs/impl/graph.py
+++ b/pacman/model/graphs/impl/graph.py
@@ -117,11 +117,6 @@ class Graph(ConstrainedObject, AbstractGraph):
         self._incoming_edges[edge.post_vertex].add(edge)
         self._outgoing_edge_partition_by_edge[edge] = partition
 
-    def get_outgoing_partition_for_edge(self, edge):
-        if edge in self._outgoing_edge_partition_by_edge:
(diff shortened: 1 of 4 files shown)


pacman/model/graphs/impl/graph.py
```diff
@@ -167,6 +171,12 @@ def outgoing_edge_partitions(self):
     def n_outgoing_edge_partitions(self):
         return len(self._outgoing_edge_partitions_by_name)
 
+    @overrides(AbstractGraph.get_outgoing_partition_for_edge)
+    def get_outgoing_partition_for_edge(self, edge):
+        if edge in self._outgoing_edge_partition_by_edge:
+            return self._outgoing_edge_partition_by_edge[edge]
+        return None
+
```
rowleya: I think that you should never be able to ask for an edge that is in a graph and find that it doesn't have an outgoing edge partition.  Therefore I would advise that this is changed to not check and just return (which will then result in a key error if the edge is not part of this graph).

alan-stokes: k

alan-stokes: done boss

PR #191, opened 2018-12-13
fixes issue with over zelious

alan-stokes:
fixes: https://github.com/SpiNNakerManchester/PACMAN/issues/190

i hope

### DIFF
--- a/pacman/operations/placer_algorithms/one_to_one_placer.py
+++ b/pacman/operations/placer_algorithms/one_to_one_placer.py
@@ -55,7 +55,9 @@ class OneToOnePlacer(RadialPlacer):
         # iterate over vertices
         for vertex_list in vertices:
             # if too many one to ones to fit on a chip, allocate individually
-            if len(vertex_list) > machine.maximum_user_cores_on_chip:
+
+            if len(vertex_list) > \
+                    machine.get_maximum_cores_available_on_a_chip():
                 for vertex in progress.over(vertex_list, False):
                     self._allocate_individual(
                         vertex, placements, resource_tracker,


rowleya:
I don't think this will fix the issue, but I will have a go at testing.  The reason I don't think this will be enough is that when the error occurs, there is a chip with enough cores available, but it seems to insist on using 0, 0 anyway.

alan-stokes:
if thats the case, its not the one to one placer thats zelious, tis the resource tracker. Im looking at its group res to see if theres something akin to that erro in there. but without an actual test to run, its a wee bit difficult.

alan-stokes:
i think its how allocate_constrained_group_resources call works. as get get_chip_and_core will return None, None, None. which means when it goes into allocate_group_resources, itll have none for chips. which means _get_usable_chips will have none which means its going to return the first chip in the list from get_chips_on_board, which obvs is 0,0. 

eventually itll go to _n_cores_available() with 0,0. 

and that code that'll be fired, i think is :

        if key in self._core_tracker:
            n_cores = len(self._core_tracker[key])
        else:
            n_cores = sum(not proc.is_monitor for proc in chip.processors)
        return n_cores - self._n_cores_preallocated[key]

i cant see how that would return more than whats available.......

you could remove the 
n_cores = sum(not proc.is_monitor for proc in chip.processors)
with 
chip.n_user_processors()

but thats just a clean up, as the chip does basically the same calc during init. 

confused.

rowleya:
Replaced by #192

rowleya:
Note - this change was essentially incorporated into the other changes.

PR #192, opened 2018-12-17
Fix virtual placement

### LINKED ISSUE #190: OneToOnePlacer might be overzealous
I have seen a circumstance under which the OneToOnePlacer might be trying a bit too hard.  The issue seemed to come up with a 512 neuron retina population connected to a single population of 256 neurons, where the connected population then has 16 atoms per core as a restriction.  This attempted to put all of the 16 cores on to 0, 0 by force.  This has two issues:
 - Nothing was constraining the population to 0, 0.  This was tested by putting a constraint of 0, 1 on the population, which then worked.
 - If OneToOnePlacer fails to place connected vertices together, it shouldn't fail but instead just place as normal.

This sounds to me like the OneToOnePlacer expecting a certain exception to occur when a grouped placement fails, but instead receiving a different one.

rowleya:
Fixes #190.  Replaces #191.

The issue turned out to be that the one-to-one placer was not ordering constraints correctly, so it was possible for it to assign vertices to a chip and then not have enough space for vertices constrained to be placed on that chip.

The fix required a fair amount of refactoring of the code, but I think it is in a better place for it, with easier to understand code (I hope).  Tests have also been added to test for the specific case, and for one-to-one placement in general.

### DIFF
--- a/pacman/operations/placer_algorithms/one_to_one_placer.py
+++ b/pacman/operations/placer_algorithms/one_to_one_placer.py
@@ -1,14 +1,17 @@
 from spinn_utilities.progress_bar import ProgressBar
-from spinn_utilities.ordered_set import OrderedSet
 from pacman.exceptions import (
     PacmanException, PacmanInvalidParameterException, PacmanValueError)
 from pacman.model.placements import Placement, Placements
 from pacman.operations.placer_algorithms import RadialPlacer
 from pacman.utilities.utility_objs import ResourceTracker
 from pacman.utilities.algorithm_utilities.placer_algorithm_utilities import (
-    get_same_chip_vertex_groups, sort_vertices_by_known_constraints)
-from pacman.model.constraints.placer_constraints import SameChipAsConstraint
-from pacman.utilities.utility_calls import is_single
+    get_same_chip_vertex_groups, get_vertices_on_same_chip, group_vertices)
+from pacman.model.constraints.placer_constraints import (
+    SameChipAsConstraint, ChipAndCoreConstraint)
+from pacman.utilities.utility_calls import (
+    is_single, locate_constraints_of_type)
+from pacman.model.graphs import AbstractVirtualVertex
+import functools
 
 
 def _conflict(x, y, post_x, post_y):
@@ -17,6 +20,62 @@ def _conflict(x, y, post_x, post_y):
     if y is not None and post_y is not None and y != post_y:
         return True
     return False
+
+
+def _find_one_to_one_vertices(vertex, graph):
+    """ Find vertices which have one to one connections with the given\
+        vertex, and where their constraints don't force them onto\
+        different chips.
+
+    :param graph: the graph to look for other one to one vertices
+    :param vertex: the vertex to use as a basis for one to one connections
+    :return: set of one to one vertices
+    """
+    # Virtual vertices can't be forced on other chips
+    if isinstance(vertex, AbstractVirtualVertex):
+        return []
+    found_vertices = []
+    vertices_seen = {vertex}
+
+    # look for one to ones leaving this vertex
+    outgoing = graph.get_edges_starting_at_vertex(vertex)
+    vertices_to_try = [
+        edge.post_vertex for edge in outgoing
+        if edge.post_vertex not in vertices_seen]
+    while vertices_to_try:
+        next_vertex = vertices_to_try.pop()
+        if next_vertex not in vertices_seen:
+            vertices_seen.add(next_vertex)
+            edges = graph.get_edges_ending_at_vertex(next_vertex)
+            if is_single(edges):
+                found_vertices.append(next_vertex)
+                outgoing = graph.get_edges_starting_at_vertex(next_vertex)
+                vertices_to_try.extend([
+                    edge.post_vertex for edge in outgoing
+                    if edge.post_vertex not in vertices_seen])
+
+    # look for one to ones entering this vertex
+    incoming = graph.get_edges_ending_at_vertex(vertex)
+    vertices_to_try = [
+        edge.pre_vertex for edge in incoming
+        if edge.pre_vertex not in vertices_seen]
+    while vertices_to_try:
+        next_vertex = vertices_to_try.pop()
+        if next_vertex not in vertices_seen:
+            vertices_seen.add(next_vertex)
+            edges = graph.get_edges_starting_at_vertex(next_vertex)
+            if is_single(edges):
+                found_vertices.append(next_vertex)
+                incoming = graph.get_edges_ending_at_vertex(next_vertex)
+                vertices_to_try.extend([
+                    edge.pre_vertex for edge in incoming
+                    if edge.pre_vertex not in vertices_seen])
+
+    extra_vertices = get_vertices_on_same_chip(vertex, graph)
+    for vertex in extra_vertices:
+        if vertex not in vertices_seen:
+            found_vertices.append(extra_vertices)
+    return found_vertices
 
 
 class OneToOnePlacer(RadialPlacer):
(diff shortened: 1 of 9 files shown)


pacman/operations/placer_algorithms/one_to_one_placer.py
```diff
@@ -35,14 +93,19 @@ def __call__(self, machine_graph, machine):
...
+
+        # Work out the vertices that should be on the same chip by one-to-one
+        # connectivity
+        one_to_one_groups = group_vertices(
+            machine_graph.vertices, functools.partial(
+                _find_one_to_one_vertices, graph=machine_graph))
 
         return self._do_allocation(
-            sorted_vertices, machine, same_chip_vertex_groups, machine_graph)
+            one_to_one_groups, same_chip_vertex_groups, machine, machine_graph)
```
alan-stokes: one_to_one_groups and same_chip_vertex_groups hold basically the same thing and have the same format. maybe its worth merging them so that its just one thing handing down to the placer code? so that its easier if there were more constraints. like sdram links

alan-stokes: this mainly as there's going to be other situations. like SDRAM link checks. maybe a list of functions should be handed down to group_vertices()? so that the common code can just handle them all????

pacman/operations/placer_algorithms/one_to_one_placer.py
```diff
@@ -52,166 +115,97 @@ def _do_allocation(
...
+        # iterate over the remaining unconstrained (or less constrained)
+        # vertices
+        for vertex in unconstrained:
+
+            # If the vertex has been placed, skip it
+            if vertex in all_vertices_placed:
+                continue
+
+            # Find vertices that are one-to-one connected to this one
+            one_to_one_vertices = one_to_one_groups[vertex]
```
alan-stokes: mmm, this makes me think its better merged, as its considered unconstrained or constrained above... based off placement (and in future sdram links) (Note for myself that branch will need a big conflict merge).

pacman/operations/placer_algorithms/one_to_one_placer.py
alan-stokes: i note that the sdram links are already in there. merged into the common code same chip checker. this makes me wonder if they should be separate ones to reflect that there's more...... unknown.

rowleya: The issue is that the one-to-one "constraint" is soft i.e. I don't have to put it on the same chip.  So if the placement of a one-to-one group can't be done, the algorithm has to fall back on trying normal placement.  However this must still respect the SDRAM edges and same-chip-as constraints.

alan-stokes: but thats reflected by the hard ones being checked by chip constraints later on. thus why you can get away with it

alan-stokes: note next comment which set me off there

rowleya: I don't think anything else in the code makes this happen though... i.e. there is nothing that checks chip constraints later on (that I know of).

alan-stokes: 119 to 126

rowleya: That only checks for ChipAndCoreConstraint.

rowleya: So this will only place vertices on the same chip if one of those vertices is further constrained.  After that, it tries one-to-one placement of all vertices, or if that fails, it does same-chip-as.

pacman/operations/placer_algorithms/one_to_one_placer.py
rowleya: Note SDRAM links already exist here (but don't have any resources associated with the edge).

alan-stokes: i noticed after typing.

pacman/operations/placer_algorithms/one_to_one_placer.py
alan-stokes: right, so the chip and core is a hard chip placement. and is done as we always done it. 

after that it only does 1 by 1 if there's no chip with enough cores. if it can do it, it does it as a group first, then if that fails, then tries with 1 by 1. 

i do agree that there's something i missed, which is with sdram links, if you fail as a group, it'll still do 1 by 1, which is wrong on a sdram link aspect...... but yours has that same issue. which kinda implies the hard constraint check is not enough, but if you got that to cover sdram links as well, you could merge them into 1 dict.

rowleya: SDRAM links are handled in the same_chip_as grouping.  The one-to-one dict includes SDRAM linked and same-chip-as linked vertices i.e. it is already merged in.  See lines 74-76.

alan-stokes: bugger, it took me a while to see that sneaky bit. But then why do lines 95? why have the second one in the first place? when the data is in first param for _do_allocation??

rowleya: You need the second one in case the first doesn't work - if you can't allocate all the one-to-one vertices on the same chip, you can still get a working placement with the same-chip-as vertices.  The one-to-one "constraint" is a soft one but the same-chip-as is not.

rowleya: Once you have them merged, you can't tell which is which.

alan-stokes: but my point is that you dont care. i think.

alan-stokes: and the issue of sdram links being split over multiple chips is still there i think

rowleya: OK, then I am telling you that you are wrong not to!  We want to try to put them all on the same chip, but if that fails, you don't need to fail completely (if you do, you are failing where another placer would have worked and that seems like a bad idea).  And if you don't fail completely, you need to make sure the actual constraints requested are respected.

rowleya: SDRAM links are being respected by the "allocate_individual" function.  If you can write a test that says they are not though, I would be willing to fix it.

alan-stokes: but your not going to fail when its the one to one constrained, youll try first on all on same chip. as line 156 ensures it wont try individual first, unless it doesnt fit on a chip, which fine, its individual. so assuming fits on chip,  itll go to line 166, which goes to 182 which does group then if that fails, (170) itll do indiviudal again. But you will not fail when your meant to when dealing with sdram links. as sdram links have no placement constraint associated with them before placer engages and so will be in weak linked, go through the same lines, at which point line 170 logic is wrong. as a sdram link work over multiple chips.

now given your so strongly in beleif im wrong, what have i missed there?

rowleya: The reason that these are not done in the same context as the ChipAndCore constraint is that they don't have a specific place to go.  So long as there is at least one chip with enough space, they can go on that chip.

alan-stokes: ya win. but ive left the test in there. least itll stop my logic path in future. 

but for my defence. line 201 " _allocate_individual(" does not do what it says. it does allocate_group. so its a bit misleading, which could be part of what took me down that road. 

and ill admit if you replace same_chip_as with one to one, it blows up, but id like to figure out why.  as i dont really get why it doesnt work. when it contains the same data in there.

rowleya: OK, to help with understanding, the sets in the one-to-one dict are super sets of those in the same-chip-as dict.  Say you have a vertex A which is one-to-one connected to vertex B which has to go on the same chip as vertex C, and one chip left with only 2 cores on it, and one with only 1 core e.g.:
A --> B <--> C  
Chip 1: 2 cores
Chip 2: 1 core

This means that the one-to-one dict has:
A: [A, B, C]
B: [A, B, C]
C: [A, B, C]

and the same-chip-as dict has:
A: [A]
B: [B, C]
C: [B, C]

Say now you try to place B, and it tries to place all three, but there isn't a chip with 3 cores, so that fails.  So you say to place vertex B "individually", which results in B and C being placed.

Does that help?

PR #195, opened 2019-01-17
removed an n * n * n problems with large one to one chains

Christian-B:


Christian-B:
Ah found and fixed one edge case that was breaking test.

The one_to_one_groups previously included the same_chip_vertex_groups so the algorithm used the  ne_to_one_groups first and fell back on the same_chip_vertex_groups if required.

As one_to_one_groups now gives stops producing very large groups in some cases it returns incorrect too small one.

Fix is to concat the two group and use the concat first.

pacman/operations/placer_algorithms/one_to_one_placer.py
```diff
@@ -86,31 +86,34 @@ class OneToOnePlacer(RadialPlacer):
...
         # Get which vertices must be placed on the same chip as another vertex
         same_chip_vertex_groups = get_same_chip_vertex_groups(machine_graph)
 
+        progress.update()
         # Work out the vertices that should be on the same chip by one-to-one
         # connectivity
         one_to_one_groups = group_vertices(
             machine_graph.vertices, functools.partial(
-                _find_one_to_one_vertices, graph=machine_graph))
+                _find_one_to_one_vertices, graph=machine_graph), 18)
```
alan-stokes: magic number

Christian-B: Agreed! 

I am going to try one alternative but I think that will be slower on machine with very large chains,

pacman/utilities/algorithm_utilities/placer_algorithm_utilities.py
```diff
@@ -63,50 +64,67 @@ def group_vertices(vertices, same_group_as_function):
...
+    return same_chip_vertices
+
+
+def add_set(all_sets, new_set):
+    for a_set in all_sets:
+        intersection = new_set & a_set
+        if intersection:
+            all_sets.remove(a_set)
+            union = a_set | new_set
+            add_set(all_sets, union)
```
dkfellows: There should be some way to avoid writing this recursively.

dkfellows: Indeed, I think there ought to be a way to write this using a single pass.

Christian-B: Yep Done

dkfellows: Looks good

pacman/operations/placer_algorithms/one_to_one_placer.py
```diff
@@ -88,31 +89,49 @@ class OneToOnePlacer(RadialPlacer):
 
     def __call__(self, machine_graph, machine):
 
+        # Iterate over vertices and generate placements
+        progress = ProgressBar(
+            machine_graph.n_vertices + 3, "Placing graph vertices")
```
alan-stokes: magic number 3. can we at least have a comment saying what these magic 3 steps are.

pacman/utilities/algorithm_utilities/placer_algorithm_utilities.py
```diff
@@ -56,7 +57,7 @@ def get_same_chip_vertex_groups(graph):
         get_vertices_on_same_chip, graph=graph))
 
 
-def group_vertices(vertices, same_group_as_function):
+def group_vertices(vertices, same_group_as_function, cutoff=sys.maxsize):
```
alan-stokes: not in python doc

pacman/utilities/algorithm_utilities/placer_algorithm_utilities.py
```diff
@@ -68,50 +69,70 @@ def group_vertices(vertices, same_group_as_function):
...
+    for group in groups:
+        for vertex in group:
+            same_chip_vertices[vertex] = group
+    for vertex in vertices:
+        if vertex not in same_chip_vertices:
+            same_chip_vertices[vertex] = {vertex}
+    return same_chip_vertices
+
+
+def add_set(all_sets, new_set):
```
alan-stokes: I might be being dense, but if the groups was a ordered set instead of a list, would this function become redundant, as you could just use update of the ordered set?

pacman/utilities/algorithm_utilities/placer_algorithm_utilities.py
```diff
@@ -68,50 +69,70 @@ def group_vertices(vertices, same_group_as_function):
...
+            same_chip_as_vertices = OrderedSet(same_chip_as_vertices)
+            same_chip_as_vertices.add(vertex)
+            # Singletons on interesting and added later if needed
+            if len(same_chip_as_vertices) > 1:
+                add_set(groups, same_chip_as_vertices)
+    return groups
 
-    return same_chip_vertices
+
+def concat_all_groups(same_chip_as_vertices, same_chip_vertices, cutoff):
```
alan-stokes: this function isnt used anywhere. redundant code

pacman/operations/placer_algorithms/one_to_one_placer.py
Christian-B: comment added

pacman/utilities/algorithm_utilities/placer_algorithm_utilities.py
Christian-B: should I reply picky or good spot? either way done.

pacman/utilities/algorithm_utilities/placer_algorithm_utilities.py
Christian-B: Added docs for the dense.

pacman/utilities/algorithm_utilities/placer_algorithm_utilities.py
alan-stokes: well here's another one. now that the ide isnt complaining about lack of python docs, its complaining that its never used....... so redundant param

pacman/utilities/algorithm_utilities/placer_algorithm_utilities.py
alan-stokes: fine, but could it be pushed into the orderedset fort encapsulation arguments? given its basically messing with sets.

pacman/utilities/algorithm_utilities/placer_algorithm_utilities.py
Christian-B: yet left over from when I was using my own function vs set builtins

Christian-B:
add_set can not be part of  orderedset  as its combines a new set into a list of existing sets.

See : PACMAN/unittests/utilities_tests/test_placer_algorithm_utilities.py for examples.

alan-stokes:
fine, was just a thought.

PR #197, opened 2019-01-30
Deal with x, y = None case properly

andrewgait:
Fix for https://github.com/SpiNNakerManchester/sPyNNaker8/issues/190.

### DIFF
--- a/pacman/model/constraints/placer_constraints/chip_and_core_constraint.py
+++ b/pacman/model/constraints/placer_constraints/chip_and_core_constraint.py
@@ -30,8 +30,8 @@ class ChipAndCoreConstraint(AbstractPlacerConstraint):
         :type p: int
         :raise None: does not raise any known exceptions
         """
-        self._x = int(x)
-        self._y = int(y)
+        self._x = None if x is None else int(x)
+        self._y = None if y is None else int(y)
         self._p = None if p is None else int(p)
 
     @property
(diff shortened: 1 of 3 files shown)


andrewgait:
The tests mentioned in that issue and those in the same directory now work on spalloc, but on a spinn-3 board the second test (https://github.com/SpiNNakerManchester/sPyNNaker8/blob/master/p8_fga_tests/test_spinnaker_link_connectors/test_spinnaker_link_2pop_different_link_ids.py) doesn't fail any more when previously it was supposed to do so (and it does fail "correctly" on spalloc, I think...).

alan-stokes:
a placement constraint which hasnt got a placement. seems like GIGO to me. How does that work as a constraint if i can set x,y,p to none?

andrewgait:
I agree, and perhaps the actual problem is that, when using spalloc, there's a call to the partitioner basically before almost anything else happens - so it creates vertices, and in this instance ApplicationSpinnakerLinkVertex.create_machine_vertex calls vertex.set_virtual_chip_coordinates (which then tries to add a constraint) before virtual_chip_x and virtual_chip_y have been set (they default to None).  So the way round it is either to deal with the None in the constraint itself, or (possibly better) to call the algorithm that sets virtual_chip_x and virtual_chip_y before the partitioner (I'm guessing this algorithm is MallocBasedChipIdAllocator, since nowhere else calls set_virtual_chip_coordinates...

andrewgait:
I could have sworn that I tried changing that yesterday evening but to no effect.  However it does seem to work after testing it using the tests in the issue, so this is ready to look at again.

PR #200, opened 2019-02-28
give different reports different names!

Christian-B:
Required by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/196 but can be merged before.


Christian-B:
There where two of each of the placement_by reports and if both where requested one would overwrite the other based on a PACMAN race condition.

PR #202, opened 2019-03-25
Commented out tag allocator check

lucaperes:
This check appears to be unnecessary and was creating a conflict in my vertex allocation


dkfellows:
Needs to be updated past the 5.0 release point…

alan-stokes:
if ive read this pr properly. that check is essential. as it says that your max neuron id is lower than a slice your trying to do. 

this all said, my work, and donals work might make this redundant.

rowleya:
I think this is now out-of-date as synapse-neuron splitting is in master...

PR #203, opened 2019-03-29
C version of router compressor with bit field

alan-stokes:
supports different types of routing table and compression

should be reviewable by now.


rowleya:
Please have a look at the branch "allow_provided_partial_tokens" and see if this fits your needs.  If so, it is simpler, and doesn't break existing tests...

PR #204, opened 2019-04-04
Fix radial placer to use constraints properly

andrewgait:
Fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/628.

@Christian-B if you want to test your script then this is the branch.


Christian-B:
This now works when merged into the Jenkins2 branch.

Will do full testing with it included now.

But it does appear to break the pacman unit tests so needs some work there.

PR #214, opened 2019-05-07
routing summary report

Christian-B:


pacman/operations/algorithm_reports/reports.py
```diff
@@ -94,6 +95,52 @@ def placer_reports_without_application_graph(
         report_folder, hostname, placements, machine)
 
 
+def router_summary_report(
+        report_folder, routing_tables,  hostname, machine):
+    """ Generates a text file of routing paths
+
+    :param routing_tables:
```
alan-stokes: fill in docs params

pacman/operations/algorithm_reports/reports_metadata.xml
```diff
@@ -116,6 +116,34 @@
...
+                <param_name>routing_tables</param_name>
+                <param_type>MemoryRoutingTables</param_type>
+            </parameter>
+            <parameter>
+                <param_name>hostname</param_name>
+                <param_type>IPAddress</param_type>
+            </parameter>
+            <parameter>
+                <param_name>machine</param_name>
+                <param_type>MemoryExtendedMachine</param_type>
```
alan-stokes: what about MemoryMachine? extended (if memory serves) is only built with external devices been added. what about a setup which never had to add any?

rowleya: The chip allocater always runs, so you always get a MemoryExtendedMachine at this point I believe.  You can of course use MemoryMachine if you are not going to look at virtual chips in any case, since they are both the same in terms of real chips.

pacman/operations/algorithm_reports/reports.py
Christian-B: Pot kettle Black!

pacman/operations/algorithm_reports/reports_metadata.xml
Christian-B: GIGO

pacman/operations/algorithm_reports/reports.py
alan-stokes: kettle pours internals on pot

pacman/operations/algorithm_reports/reports_metadata.xml
alan-stokes: so can have a second type of MemoryMachine (and it might even be better as just a MemoryMachine, as rowley hath pointed out). so not GIGO.  😛

Christian-B: copied directly from routingInfoReports   so if this wrong it is because existing code was wrong so GIGO!

alan-stokes: so your telling me that coz the foundations is wrong, your just going to keep building on the shoddy foundations? remind me never to go to a house you built. 😜

PR #216, opened 2019-05-10
fake machone even for version None and minor speedup

Christian-B:
We need a fake machine whenever max x or may y is greater than 7 no matter what the version is.

Oh and lets not do + 1  96 times per board when we can do twice.


rowleya:
The lines changed in this branch appear to have been removed from master... assuming it is therefore not needed, but will leave in case it is.

Christian-B:
no longer needed.

PR #218, opened 2019-05-16
Spreader placer

alan-stokes:
new placer. spreads vert over chips to ensure minimal incoming packets per chip and starts from midpoint of machine

does not solve roberts problem. exploring why. 

could help with @Christian-B key allocator


alan-stokes:
master merge and still working. shows significant imporvement to random balanced networks doing this. still attempting to get data on improvements on column

rowleya:
This is not enabled by default, and it seems to work and passes tests, so I am merging it.  We will need another PR if this proves to be the better algorithm and we want to make it the default.

PR #219, opened 2019-05-20
Compression

Christian-B:
Early prototype but not used by default so little harm in merging.


rowleya:
Deleting this as now merged compressor2 which has the same (but additional) functionality.

PR #220, opened 2019-05-29
Compression2

Christian-B:
Builds on https://github.com/SpiNNakerManchester/PACMAN/pull/219

Adds a tool that where possible does zoned Routing table entries as well.


rowleya:
This works fine as tested by others.  Another PR will be needed to enable this by default if this is deemed necessary.

PR #229, opened 2019-06-12
Router json

Christian-B:
Depends on https://github.com/SpiNNakerManchester/SpiNNMachine/pull/112 

Allow the uncompressed router table to be save in json to the reports folder.

Could later be reused for comprersion testing

Works with any RoutingInfoAllocator  and can be done boefre on Chip or on host compression.


Christian-B:
This functionality is turned on by adding ConvertToJsonRoutingTables to cfg section [Mapping] value: machine_graph_to_machine_algorithms and/or machine_graph_to_virtual_machine_algorithms

PR #230, opened 2019-06-20
Unordered compressor

Christian-B:
Depends on https://github.com/SpiNNakerManchester/SpiNNMachine/pull/112

Extends on https://github.com/SpiNNakerManchester/PACMAN/pull/229

Two options to run this code.
1. for actual machines in cfg section [mapping] value machine_graph_to_machine_algorithms
add ,UnorderedPreCompressor at the end

2. for actual machines in cfg section [mapping] value machine_graph_to_machine_algorithms
add ,PairCompressor at the end
AND: from loading_algorithms remove MundyOnChipRouterCompression
Which will probably be done by adding to your local cfg
loading_algorithms = None

3. For virtual machines value machine_graph_to_virtual_machine_algorithms
add UnorderedPreCompressor before the current compressor or replace the current compressor with
PairCompressor

Note: Pair compressor is the same algorithm but different code to that tested by @SimonDavidson 

there is also an independant C version of the pair compressor 
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/465


Christian-B:
This has been updated to how I intend to do it in C

Christian-B:
Does NOT need to wait for https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/458

PR #232, opened 2019-07-11
bits needed for sdram edges

alan-stokes:
depends with https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/455

### DIFF
--- a/pacman/model/graphs/machine/machine_sdram_edge.py
+++ b/pacman/model/graphs/machine/machine_sdram_edge.py
@@ -1,0 +1,36 @@
+from pacman.model.graphs.common import EdgeTrafficType
+from pacman.model.graphs.machine import MachineEdge
+
+
+class SDRAMMachineEdge(MachineEdge):
+
+    __slots__ = [
+        "_sdram_size",
+        "_sdram_base_address"
+
+    ]
+
+    def __init__(self, pre_vertex, post_vertex, sdram_size, label):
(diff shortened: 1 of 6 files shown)


pacman/model/graphs/machine/machine_sdram_edge.py
```diff
@@ -0,0 +1,36 @@
...
+            label=label, traffic_weight=1)
+        self._sdram_size = sdram_size
+        self._sdram_base_address = None
+
+    @property
+    def sdram_size(self):
+        return self._sdram_size
+
+    @property
+    def sdram_base_address(self):
```
rowleya: We may instead want an SDRAMAllocation object, similar to the Placement object

alan-stokes:
so this is dead. i think the changes are the same. but they been cherry picked from roberts changes in 
https://github.com/orgs/SpiNNakerManchester/projects/22

so this pr and project getting killed for clarity

pacman/model/graphs/machine/machine_sdram_edge.py
alan-stokes: i dont agree. your adding complexity that isnt needed. the partition internally is doing said allocation based off its type. making a allocation obejct for this will just muddy the water

alan-stokes: p.s. sorry that this discussion is being done in a dead branch.

PR #235, opened 2019-07-17
fixes chimps lazyness

alan-stokes:


rowleya:
> I disagree!
> 
> From within the same class I think it is perfectly acceptible to access parameters directly!

This is correct, though note that if you made the variables double-underscore, I don't think you can (i.e. private variables don't work the same as they do in Java).  If we aren't going to do this, they can be left as they are.  As a side note, the "else" should then probably be "else if isinstance(other, VariableSDRAM) to protect the access from other rubbish, but then equally we don't do this sort of check anywhere else i.e. the assumption is that the input here is an SDRAM object.

alan-stokes:
well if your going to reject my fix. then you might as well delete the brnach as well

Christian-B:
I was being polite and leaving the branch for you to argue your case.

Now that I have won the branch has been deleted.

alan-stokes:
think you win only coz the energy levels have dropped below the effort to reach my cliff. lol

PR #240, opened 2019-08-09
Sorted zoned

Christian-B:
Zonded infio allocator version 2.

Handles machine vertexes partitions with multiple partitions.
Note: ApplicationRoutingInfos is not yet used by any other algorithms so left for farther development 

Handles FixedKeyAndMaskConstraint. Even ones with a VERY big number of neurons/keys needed

If multiple application vertexes have exactly the same set of target cores they will be infos


alan-stokes:
conflicts

Christian-B:
Too far behind and too messy to merge!

alan-stokes:
deleting branch or not?

Christian-B:
branch left of history

PR #241, opened 2019-08-09
FixedRouteRouter does not need to be complicated!

Christian-B:
A simple water drop algorithm.

Works even with one way links.


andrewgait:
Would it be better if we kept FixedRouteRouter as it was and renamed this to BasicFixedRouteRouter?

Christian-B:
I see no reason to keep 2 different routers here as there is no use cases where the previous version is better than this one.

alan-stokes:
given the original was designed when we thought we could have multiple streams going into the SUPG to max out the Ethernet, but then we found 1 stream did that. so all the intelligence was not needed, but at that point this had been done and was working. so im happy for it to be simplified now that we have a better idea of the speed up limitations

PR #245, opened 2019-08-13
Fix the divide by zero problem for real this time

dkfellows:
#243 didn't do the trick, but this does…


dkfellows:
Good. Integration tests pass.

PR #247, opened 2019-08-14
Diamonds are not forever

Christian-B:
AbstractVirtualVertex, AbstractSpiNNakerLinkVertex and AbstractFPGAVertex where only ever used to check the instance of an Object that due to context was KNOWN to be an AbstractVertex.

ApplicationSpiNNakerLinkVertex, ApplicationFPGAVertex, MachineSpiNNakerLinkVertex and MachineFPGAVertex had diamond inheritance to AbstractVertex through  ApplicationVertex or MachineVertex 

All other uses of AbstractVirtualVertex build through one of the Application/MaichineSpiNNaker/FPGALinkVertex classes so get AbstractVertex that way.

This PR removes the AbstractVertex from AbstractVirtualVertex super list and renames it AbstractVirtual to show the change.

Removing the Diamond inheritance allows a none empty slots to be used in AbstractVertex.
It also allows AbstractVertex to have ConstrainedObject as its super 

This PR is farther developed in https://github.com/SpiNNakerManchester/PACMAN/pull/246
Which moves the methods and slots down.


dkfellows:
Shirley Bassey will be so upset.

PR #256, opened 2019-09-04
Graph mapping sanity

dkfellows:
Greatly cleans up what we do with graphs and graph mappers.

In particular, application vertices know (because they manufacture) their machine vertices, and each machine vertex knows what application vertex made it (if any; `None` is legal there when dealing with a pure machine graph). Similarly for edges. This makes the graph mapper redundant; I've left it in just in case someone makes one explicitly, but none of our code does anything meaningful with it any more.


pacman/executor/algorithm_metadata_xml_reader.py
```diff
@@ -51,45 +48,61 @@ class AlgorithmMetadataXmlReader(object):
 
     __slots__ = [
         # paths to extra metadata files
-        "_xml_paths"
+        "_xml_paths",
+        # current path
+        "_xml_path"
```
Christian-B: I dont like making one of many parameters a self object.

Permission to change?

pacman/model/graphs/graph.py
```diff
@@ -56,28 +50,26 @@ class Graph(ConstrainedObject, AbstractGraph):
...
-        :param allowed_partition_types:\
-            A single or tuple of types of partitions to be allowed in the graph
+        :type allowed_edge_types: type or tuple(type, ...)
         :param label: The label on the graph, or None
+        :type label: str or None
         """
         super(Graph, self).__init__(None)
         self._allowed_vertex_types = allowed_vertex_types
         self._allowed_edge_types = allowed_edge_types
-        self._allowed_partition_types = allowed_partition_types
```
Christian-B: While this change makes sense in the current stack it does not when considering the future SDRAM edges stuff.

  Will revert in the Cherry pick

requirements.txt
```diff
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 six
+Deprecated
```
Christian-B: Why?

pacman/model/graphs/machine/machine_vertex.py
```diff
@@ -16,28 +16,88 @@
...
             label = str(type(self))
         super(MachineVertex, self).__init__(label, constraints)
         self._added_to_graph = False
+        self._app_vertex = app_vertex
+        self._index = None
+        if app_vertex is not None and not isinstance(
+                app_vertex, ApplicationVertex):
+            raise PacmanInvalidParameterException(
+                "app_vertex", app_vertex, "must be an application vertex")
+        if vertex_slice is not None:
```
Christian-B, line 60: I think remember_associated_machine_vertex should be called from here rather than in different places.

pacman/operations/partition_algorithms/basic_partitioner.py
```diff
@@ -116,7 +119,7 @@ def _partition_one_application_vertex(
                 "{}:{}:{}".format(vertex.label, first, last),
                 get_remaining_constraints(vertex))
             m_graph.add_vertex(m_vertex)
-            mapper.add_vertex_mapping(m_vertex, vertex_slice, vertex)
+            vertex.remember_associated_machine_vertex(m_vertex)
```
Christian-B, line 122: I wonder if it would not be better to have Machine_vertex init do the registrastion.

Otherwise it is sometimes in the partitioner and sometimes in create_machine_vertex (for example LivePacketGather)

pacman/executor/algorithm_metadata_xml_reader.py
dkfellows: I do make use of that assumption (in many places!), and it greatly reduces the number of variables that get passed around explicitly through lots of calls just because some leaf method wants to look something up or make a simple call; such things I think of as “context” and those I like to put hanging off `self` as that's the standard context thing in Python.

pacman/model/graphs/graph.py
dkfellows: I think I put this back in.

requirements.txt
dkfellows: Unneeded.

dkfellows:
My general comment is that I want an `ApplicationFoo` to be the _manufacturer_ of all `MachineFoo`s that it is associated with. I've gone this way particularly with vertices, but it properly applies to edges and graphs as well, at least in my mind.

It does suggest that moving on from this to self-partitioning (not in this PR!) is a good plan.

pacman/model/graphs/application/application_edge.py
```diff
@@ -14,9 +14,19 @@
...
 from pacman.model.graphs.common import EdgeTrafficType
-from pacman.model.graphs.machine import MachineEdge
+
+_MachineEdge = None
+
+
+def _machine_edge_class():
+    global _MachineEdge  # pylint: disable=global-statement
+    if _MachineEdge is None:
+        from pacman.model.graphs.machine import MachineEdge
```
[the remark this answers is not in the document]
dkfellows: Somewhere we make a mess of the declaration order Directed Acyclic Graph. I don't know where.

pacman/model/graphs/application/application_edge.py
```diff
@@ -61,27 +72,35 @@ def __init__(
         self._pre_vertex = pre_vertex
         self._post_vertex = post_vertex
         self._traffic_type = traffic_type
+        if machine_edge_type is None:
+            machine_edge_type = _machine_edge_class()
```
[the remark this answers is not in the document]
dkfellows: This would be just `MachineEdge` (i.e., effectively a constant as it is a class reference) except I want the resolution to be done at a later time. I put that lookup in its own function simply because it is a nasty bit of ghastliness and so should be corralled on its lonesome.

pacman/model/graphs/application/application_edge.py
```diff
@@ -61,27 +72,35 @@ def __init__(
...
+        # TODO: Enforce the type relationship
         self._machine_edge_type = machine_edge_type
+        self.__machine_edges = OrderedSet()
 
     @property
     @overrides(AbstractEdge.label)
     def label(self):
         return self._label
 
+    # This method gets overridden
```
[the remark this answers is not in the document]
dkfellows: It's not that it _must_ happen, but rather that overriding is required if you want to pass alternate arguments to the machine edge subclass constructor.

Refactored for clarity

pacman/model/graphs/machine/machine_graph.py
```diff
@@ -15,20 +15,36 @@
...
 
 class MachineGraph(Graph):
     """ A graph whose vertices can fit on the chips of a machine.
     """
 
-    __slots__ = []
+    __slots__ = ["_app_graph"]
 
-    def __init__(self, label):
+    def __init__(self, label, application_graph=None):
```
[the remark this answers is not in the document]
dkfellows: Have to be a little bit careful here because of graph filtering (in sPyNNaker right now, but probably should be in FEC now that I've put the filtering rules themselves in the right place); the ordering of operations there is very precise.

pacman/model/graphs/application/application_graph.py
```diff
@@ -15,21 +15,49 @@
...
+
+    @property
+    def machine_graph(self):
+        """
+        :rtype: MachineGraph
+        """
+        return self.__machine_graph
+
+    @machine_graph.setter
+    def machine_graph(self, machine_graph):
```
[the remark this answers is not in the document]
dkfellows: As noted, this is something that one has to be careful of. However I agree in principle.

pacman/model/graphs/graph.py
```diff
@@ -208,12 +207,11 @@ def add_outgoing_edge_partition(self, outgoing_edge_partition):
...
         """
         # verify that this partition is suitable for this graph
-        if not isinstance(
-                outgoing_edge_partition, self._allowed_partition_types):
+        if not isinstance(outgoing_edge_partition, OutgoingEdgePartition):
             raise PacmanInvalidParameterException(
                 "outgoing_edge_partition", outgoing_edge_partition.__class__,
                 "Partitions of this graph must be one of the following types:"
-                " {}".format(self._allowed_partition_types))
+                " {}".format(OutgoingEdgePartition.__class__))
```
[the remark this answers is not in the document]
dkfellows: You're right. Restored `_allowed_partition_types` in many places.

pacman/model/graphs/machine/machine_edge.py
```diff
@@ -52,16 +55,23 @@ def __init__(
...
         self._pre_vertex = pre_vertex
         self._post_vertex = post_vertex
         self._traffic_type = traffic_type
         self._traffic_weight = traffic_weight
+        self._app_edge = app_edge
 
     @property
     @overrides(AbstractEdge.label)
     def label(self):
+        """
```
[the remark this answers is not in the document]
dkfellows: fair enough

pacman/model/graphs/machine/machine_vertex.py
```diff
@@ -16,28 +16,88 @@
...
+        """
+        return self._index if self._index is not None else 0
+
+    @index.setter
+    def index(self, value):
+        """ The index into the collection of machine vertices for an
+            application vertex.
+        """
+        if self._index is not None:
+            # TODO: Should we error or warn about this?
```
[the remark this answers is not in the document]
dkfellows: ok

pacman/model/graphs/machine/machine_vertex.py
dkfellows: If that works with doing graph filtering, yes.

pacman/operations/partition_algorithms/basic_partitioner.py
dkfellows: Or rather that you can do it when a machine vertex is added to its graph provided the machine vertex knows about its originating application vertex. Which it should do; it was made by it.

pacman/model/graphs/machine/machine_vertex.py
dkfellows: I've instead made it be done when the vertex is added to the graph. There was only one case where it wasn't previously the step immediately after adding to the graph, and that was a case (LPG machine vertex creation) where changing to be that way was fine.

pacman/model/graphs/machine/machine_graph.py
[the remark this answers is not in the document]
dkfellows: The reason why I allow `None` is because of the graph front end. I do not wish to change that API.

PR #261, opened 2019-10-03
Make vertices set again

rowleya:
Make vertices a set to make adding edges quicker


rowleya:
This speeds up the partitioning operation of one example from over a week (after which I gave up) to about 1 hour.

PR #265, opened 2019-10-24
Try to make the 1-to-1 placer faster

dkfellows:
I overheard that the one-to-one placer is unacceptably slow, so I had a look at it to see if there were any reasonably simple fixes that could be done. There were a few.


rowleya:
This does appear to make it faster (took about 9 hours after vs. 24 hours before), so merging.

PR #269, opened 2019-11-08
better error meesgae on core constraint

Christian-B:
gives a better error message for when a core constraint is not possible:

As much as I plan to do for https://github.com/SpiNNakerManchester/PACMAN/issues/268

### DIFF
--- a/pacman/utilities/utility_objs/resource_tracker.py
+++ b/pacman/utilities/utility_objs/resource_tracker.py
@@ -1225,6 +1225,16 @@ class ResourceTracker(object):
                         reverse_ip_tags_allocated)
 
         # If no chip is available, raise an exception
+        if chips is not None and processor_id is not None:
+            if len(chips) == 1:
+                (x, y) = chips[0]
+                raise PacmanValueError(
+                    "Core {}:{}:{} is not available.".format(
+                        x, y, processor_id))
+            else:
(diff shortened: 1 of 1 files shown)


alan-stokes:
failed tests

pacman/utilities/utility_objs/resource_tracker.py
```diff
@@ -1233,7 +1233,7 @@ def allocate_resources(self, resources, chips=None,
                         x, y, processor_id))
             else:
                 raise PacmanValueError(
-                    "Processor id is not available on any of the chips"
+                    "Processor id {} is not available on any of the chips"
```
alan-stokes: dont you also need to define the chips. given chip constraints limit it to 1. whereas any chips is all.

Christian-B: The line above deals with the typical case of core constraint as then len(chips) == 1.

Not sure if this line will every be hit but put it in just in case.

alan-stokes: yes i agree that your any chip list is 1. but the error message imples over all the machine

Christian-B: if len(chips) == 1: ELSE this error

PR #272, opened 2019-11-28
Timestep changes 

Christian-B:
This is an alternative solution for different timesteps per vertex.

Before this PR is merged we need to decide.
1. Do we at the python level wish to support multiple timesteps.
2. Do we want to support Application Vertexes with more than one timestep.

The current state of the PR is to support both 1 and 2 as yes.
I @Christian-B prefer to force an ApplicationVertex to have just a single timestep which would require a cleanup here before merging.

https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/528
https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/134
https://github.com/SpiNNakerManchester/lab_answers/pull/4
https://github.com/SpiNNakerManchester/sPyNNaker/pull/749
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/317
https://github.com/SpiNNakerManchester/sPyNNaker8NewModelTemplate/pull/48

This effects a HUGH amount of stuff so I wonder if having different "timestep"s for different vertexes is a good idea.
An Alternative is to have only a single Python timestep which is used EVERYWHERE with the  C code of the vertexes running at a different speed doing any conversions. 
With possibly minimum python changes to make sure the runtime is rounded up to a lowestCommonMultiple of the C runtimesteps.


Christian-B:
Too many changes too complex to be worth it,
Leaving branch for histocial reasons.

PR #273, opened 2019-12-09
Remove Broken Flexi-Fields

dkfellows:
They didn't work. None of our routing info allocators supported them. They were untested. They were unused.

They are now gone.


alan-stokes:
fine by me. if we ever decide to revisit that complicated brain path, its likely we'd need to rebuild from scratch anyhow

PR #279, opened 2020-02-17
makes things cleaner

alan-stokes:
cleaning


alan-stokes:
fixed

Christian-B:
Overkill to pass parameters all over the place when the algorithm is only designed to run once per Object.

But Alan had a bad day yesterday so not worth the fight.

alan-stokes:
overkill vs a pep8 magic thing to allow you to create self params outside of the init, or the call?! thats just bad coding. Bad day!?

PR #283, opened 2020-02-24
cherries from https://github.com/SpiNNakerManchester/PACMAN/pull/256

Christian-B:


dkfellows:
Do you want to add more to these?

PR #284, opened 2020-02-25
Graph fixes

Christian-B:
Changes from https://github.com/SpiNNakerManchester/PACMAN/pull/256 

However allowed_partition_types put back in 

this will cause clashes with https://github.com/SpiNNakerManchester/PACMAN/pull/256  sorry,


dkfellows:
> this will cause clashes with #256 sorry

Don't be sorry. I already told you I don't mind!

PR #287, opened 2020-03-03
Json graphs

Christian-B:
Writes a partial description of the application and machine graphs so  that together with json_partition_n_keys_map it has enough info to rerun partitioning, key allocation, routing and router compression

Requires: https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/559

https://github.com/SpiNNakerManchester/sPyNNaker/pull/796 allows n keys to come out earlier.

tested with https://github.com/SpiNNakerManchester/sPyNNaker8/pull/368


pacman/utilities/json_utils.py
```diff
@@ -283,57 +335,125 @@ def vertex_add_contstraints_from_json(json_dict, graph):
...
+            json_list.append(application_vertex_to_json(
+                application_vertex, graph_mapper))
+        json_dict["vertices"] = json_list
+        json_list = []
+        for partition in machine_graph.outgoing_edge_partitions:
+            json_list.append(partition_to_json(partition))
+        json_dict["partitions"] = json_list
+    except Exception as ex:  # pylint: disable=broad-except
+        json_dict["exception"] = str(ex)
+    return json_dict
```
rowleya: This seems to take the vertices from the application graph but the edges from the machine graph.  Is this what you wanted?

pacman/utilities/json_utils.py
```diff
@@ -283,57 +335,125 @@ def vertex_add_contstraints_from_json(json_dict, graph):
...
+                machine_vertex, application_vertex)
+    # Only do constraints when we have all the vertexes to link to
+    for j_application_vertex in json_dict["vertices"]:
+        vertex_add_contstraints_from_json(
+            j_application_vertex, application_graph)
+        for j_machine_vertex in j_application_vertex["machine_vertices"]:
+            vertex_add_contstraints_from_json(j_machine_vertex, machine_graph)
+    for j_partitions in json_dict["partitions"]:
+        partition_from_json(j_partitions, machine_graph)
+    return application_graph, machine_graph, graph_mapper
```
rowleya: So far as I can tell, this never adds an edge to the application graph, so it won't be the application graph that went in.  Is this what you wanted?

pacman/utilities/json_utils.py
Christian-B: Yes The only role of the Application vertexes is to group the Machine Vertexes by Application vertex so that the key Allocator and ensure the keys for two different Application Vertexes do not intersect.

pacman/utilities/json_utils.py
Christian-B: To the best of my knowledge once Machine Edges are created none of the placers, routers, key allocators and their support algorithms ever use Application Edges.

rowleya: I don't think any of the current ones do; note though that if we come up with one that can make use of it, this information will be lost.  If you are happy with "adding this when needed", I am happy to merge this, just so long as we understand what is missing.

Christian-B:
I currently have some very large data generated by @pabogdan Will now test this to see if it worked.

rowleya:
Is this still relevant?  I think something similar hit master recently...

rowleya:
OK, it looks like this does still have some changes that are not in master.  I am happy enough to merge these changes without those in SpiNNakerManchester/sPyNNaker#796, or else we can make the fix needed there first...

Christian-B:
Lets wait for View and reduce overhead and then see what we need

Christian-B:
Some much has changed probably easier to redo if ever needed.

PR #293, opened 2020-03-06
Multi-region SDRAM

rowleya:
An idea for allow multi-region SDRAM.  There are some issues in this implementation that might need some though:
 1. It doesn't support subtract (do we actually do this anywhere though?)
 1. Addition is done by adding regions.  This may or may not make sense depending on the context.  If just a total is being expected at the end, this is fine...
 1. Regions are just a list; should they be more complex e.g. with a region id?

The idea of this is that if a vertex returns one of these in the SDRAM, the DSG and/or DSE can check the region sizes against the SDRAM resources provided and make sure no regions are going over what they said.  An extension of this would be that the DSG could allocate the regions up-front which would further enforce this.

### DIFF
--- a/pacman/model/resources/multi_region_sdram.py
+++ b/pacman/model/resources/multi_region_sdram.py
@@ -13,7 +13,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 from .abstract_sdram import AbstractSDRAM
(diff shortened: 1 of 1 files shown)


Christian-B:
This PR has been replaced with https://github.com/SpiNNakerManchester/PACMAN/pull/294

Thank you for the great idea.

alan-stokes:
is there a reason for keeping this brnach alive?

PR #294, opened 2020-03-16
Multi Region Sdram 

Christian-B:
Inspired by https://github.com/SpiNNakerManchester/PACMAN/pull/293

This provides a working of a dictionary based MultiRegionSDRAM
Allows the sdram cost to be assigned to specific causes and even nested
Unless the details are needed this acts like a normal VariableSDRAM

Different MultiRegionSDRAM can be combined using merge(..) which retains the dictionary.
Or even nested using the nest() method

However when doing add or sub the dictionary information is not retained as it is assumed that at a gobal level this is not interesting.

Added a report method to all sdram types so they can print/write to file in the most suitable way.

This PR is reguired for but can be merged before later PRs.

### DIFF
--- a/pacman/model/resources/multi_region_sdram.py
+++ b/pacman/model/resources/multi_region_sdram.py
@@ -115,7 +115,8 @@ class MultiRegionSDRAM(VariableSDRAM):
 
     @overrides(AbstractSDRAM.report)
     def report(self, timesteps, indent="", preamble="", target=None):
-        super(MultiRegionSDRAM, self).report(timesteps, indent, preamble, target)
+        super(MultiRegionSDRAM, self).report(
+            timesteps, indent, preamble, target)
         for region in self.__regions:
             self.__regions[region].report(
                 timesteps, indent+"    ", str(region)+":", target)
--- a/pacman/model/resources/multi_region_sdram.py
+++ b/pacman/model/resources/multi_region_sdram.py
@@ -12,10 +12,8 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-from .abstract_sdram import AbstractSDRAM
 from .constant_sdram import ConstantSDRAM
 from .variable_sdram import VariableSDRAM
-from spinn_utilities.overrides import overrides
 
 
 class MultiRegionSDRAM(VariableSDRAM):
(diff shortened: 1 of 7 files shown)


pacman/model/resources/multi_region_sdram.py
```diff
@@ -0,0 +1,109 @@
...
+        if per_timestep_sdram:
+            sdram = VariableSDRAM(fixed_sdram, per_timestep_sdram)
+        else:
+            sdram = ConstantSDRAM(fixed_sdram)
+        if region in self.__regions:
+            self.__regions[region] += sdram
+        else:
+            self.__regions[region] = sdram
+
+    def nest(self, region, other):
```
rowleya: It would be good to describe what this does

Christian-B: done

Christian-B:
Here is an example of it in use:
 core (0,0,1) Fixed 856 bytes Per_timestep 92 bytes for a total of 123469732
     recording_channel_t: Constant 128 bytes
     v: Fixed 0 bytes Per_timestep 28 bytes for a total of 37577484
     recording malloc: Constant 32 bytes
     gsyn_exc: Fixed 0 bytes Per_timestep 28 bytes for a total of 37577484
     gsyn_inh: Fixed 0 bytes Per_timestep 28 bytes for a total of 37577484
     spikes: Fixed 0 bytes Per_timestep 8 bytes for a total of 10736424
     Data_Spec_size: Fixed 688 bytes Per_timestep 0 bytes for a total of 688
         POPULATION_BASED_REGIONS.SYSTEM: Constant 12 bytes
         POPULATION_BASED_REGIONS.NEURON_RECORDING: Constant 532 bytes
         POPULATION_BASED_REGIONS.PROVENANCE_DATA: Constant 40 bytes
         POPULATION_BASED_REGIONS.SYNAPSE_PARAMS: Constant 8 bytes
         POPULATION_BASED_REGIONS.SYNAPSE_DYNAMICS: Constant 0 bytes
         SYNAPTIC_MATRIX/DIRECT_MATRIX: Constant 12 bytes
         POPULATION_BASED_REGIONS.POPULATION_TABLE: Constant 8 bytes
         POPULATION_BASED_REGIONS.CONNECTOR_BUILDER: Constant 0 bytes
         POPULATION_BASED_REGIONS.PROFILING: Constant 4 bytes
         dsg overhead: Constant 72 bytes
     DSG_malloc: Constant 8 bytes

This contains numbers that are directly comparable to other places.

rowleya:
This looks good, and I look forward to making use of it!  Of note a possible idea would be to have a DSGRegion class of some sort that indicates that the space being marked corresponds to a region allocated with DSG.  This could then be automatically checked that it matches with the allocation when finally done, and either warned or error raised if not.  This is for a future PR however.

PR #297, opened 2020-06-10
Multi sdram

Christian-B:
Adds sdram with automatic malloc


dkfellows:
I'm not sure that `DsSDRAM` is the best of names, but I've not got any better suggestions that aren't insanely wordy.

Christian-B:
closing as very old so clearly not needed.

PR #300, opened 2020-07-23
makes partitioning report go in right place

alan-stokes:
another graph mapper legacy issue


alan-stokes:
ermmm no idea what we want to do with this. given its used by deterism (i think its a actual renamed branch for deterism)

andrewgait:
> ermmm no idea what we want to do with this. given its used by deterism (i think its a actual renamed branch for deterism)

It looks fine to me.  Not sure why it didn't get merged earlier... ?  Happy to merge it now if it's not in another branch already.

alan-stokes:
i think if it gets merged, then itll close the determinism version of it. so go ahead if your happy. one less branch to keep track of

PR #302, opened 2020-07-30
Documentation polishing

dkfellows:
Fixing documentation and setting it so that we can enable auto-publishing to github pages if we want. (It's still gated/controlled by an access control token in the Travis configuration that's not on github.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,17 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 language: python
-python:
-  - 2.7
-  - 3.6
-  - 3.7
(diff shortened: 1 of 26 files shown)


dkfellows:
This should fix all the doc problems except for those addressed in #297

dkfellows:
Enabling publishing to github pages is a matter of setting the `GITHUB_TOKEN` environment variable for builds of the `master` branch using the Travis configuration. I have a valid token for that if/when we decide to turn it on.

dkfellows:
This is now converted to using Github Actions

dkfellows:
Massively out of date.

PR #306, opened 2020-09-08
Determism

alan-stokes:
ermmm not sure whats here isnt in something else as well. but was trying to get deterism up to date and passing tests again.


alan-stokes:
ermmmm im not sure why this still thinks theres changes. but given pycharm says its only line seperators differences in conf.py im going to kill this one. 1 less pr to worry about for determism.

PR #308, opened 2020-10-02
Pre vertexes by app and partition name

Christian-B:
This pr has two distinct but interwoven changes,

1. The remember_associated_machine_vertex is now triggered by the machineGraph
This assures that the set of machine vertices you get by asking all app vertices in the app_graph for their machine_vertices is the same as asking the machine_graph for all machine_vertices.

The Machine Graph assures that all or none of the machine vertices have an app vertex.

2. This changed the group key allocation to be ApplicationVertex and Partition Identifiers based.
This to provide better support when splitter result in multiple partitions  associated with an ApplicationVertex,

In most cases the Pair ApplicationVertex and Partition Identifiers  will equate to one Application Level Partition.

The data is however generated from the MachineGraph and using the Machine Level Partition Identifier which allows splitters/partitioners the right to not exactly map Application level Partitions and Machine level ones.
It also allow the addition of internal machine edges/partitions with a need to have  matching application ones.

The machine prevertex in machine edge must have an ApplicationVertex so the zoned currently do not work for only a machine level. TO BE FIXED.

The Algorithms still can not handle the GFE where vertexes are added directly to the MachineGraph but then they could not in the past either.

The Global approach to key zones is not done by the same code.  It is now more flexible as it can handle the Retina while still leaving as many other vertexes as possible with global zones.

Must be done at the same time as 
https://github.com/SpiNNakerManchester/sPyNNaker/pull/898
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/675

Added Integration tests in https://github.com/SpiNNakerManchester/sPyNNaker8/pull/472


pacman/utilities/json_utils.py
```diff
@@ -322,6 +322,7 @@ def graph_to_json(graph):
 
 def graph_from_json(json_dict):
     json_dict = json_to_object(json_dict)
+    # TODO App level if exists needed
```
alan-stokes, line 325: new TODO. was that intentional. or were you going to do something and forgot?

alan-stokes (changes requested):

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -180,3 +245,48 @@ def __bits_needed(size):
...
+    :rtype: tuple(RoutingInfo,
+        dict((ApplicationVertex, str), BaseKeyAndMask))
+    :raise PacmanRouteInfoAllocationException:
+    """
+    # check that this algorithm supports the constraints put onto the
+    # partitions
+
+    allocator = ZonedRoutingInfoAllocator(machine_graph, n_keys_map, False)
+
+    allocator._calculate_zones()
```
alan-stokes, line 290: dislike these accessing protected variables.

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -180,3 +245,48 @@ def __bits_needed(size):
...
+    :rtype: tuple(RoutingInfo,
+        dict((ApplicationVertex, str), BaseKeyAndMask))
+    :raise PacmanRouteInfoAllocationException:
+    """
+    # check that this algorithm supports the constraints put onto the
+    # partitions
+
+    allocator = ZonedRoutingInfoAllocator(machine_graph, n_keys_map, True)
+
+    allocator._calculate_zones()
```
alan-stokes, line 269: same issue

pacman/utilities/json_utils.py
Christian-B: Intentional todo to makr out of scope area.

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: They are intentionally protected as they should not be called directly.

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: There are two ways of doing it that differ so  minorly that just passing in a different flag tot he init is enough.

Pacman execution engine does not support changing param nor static methods but it does support file level methods which can then add the extra flag.

Christian-B:
To be split into multiple branch. Please do not delete branch until that is done.

PR #309, opened 2020-10-06
Get the abstract splitter slice to act like basic paritioner

Christian-B:
The SplitterFixedSliceSized tried to do what previously the basic paritioner did.

The issue was that it ONLY worked for Legacy API as would require a superclass for EVERY vertex class that does not support Legacy.

This PR allows the AbstractSplitterSlice to act like the basic paritioner did.
Which automatically allows SplitterAbstractPopulationVertexSlice and SplitterDelayVertexSlice to do so to.

The ideal would be that if request by some flag (cfg setting ect) the user could turn off all apadtive splitting. 

I aproved will add the param to SplitterAbstractPopulationVertexSlice and SplitterDelayVertexSlice so users can turn these of adaptive on individual Populations.

Not sure why this is useful but then not sure why BasicPartitioner was either.


alan-stokes:
so does this still count under new way?

Christian-B:
Closed as basic paritioner styple paritition currently not planned.


Please leave branch just in case

PR #310, opened 2020-10-08
Disable same atoms as vertex

Christian-B:
SameAtomsAsVertexConstraint has been disabled.

it will throw an NotImplementedError is is unusable!

Left the code in for now in until the whole project is in active use.

WARNING:  DelayExtensionVertex was still adding this constraint last time I checked.


alan-stokes:
so assuming self partitioning passes. this merge would solve it. when the dleay extension has it removed?

Christian-B:
> so assuming self partitioning passes. this merge would solve it. when the dleay extension has it removed?

Yes I think so

pacman/model/constraints/partitioner_constraints/same_atoms_as_vertex_constraint.py
```diff
@@ -31,6 +31,10 @@ def __init__(self, vertex):
         :param ApplicationVertex vertex:
             The vertex to which the constraint refers
         """
+        raise NotImplementedError(
+            "SameAtomsAsVertexConstraint is no longer supported. "
+            "Please contact spinnakerusers@googlegroups.com to discuss your "
+            "requirements.")
         self._vertex = vertex
```
alan-stokes, line 38: unreachable code

Christian-B: I sure hope it is unreachable from OUR code otherwise we would have a problem!

As we are unsure if any user was using this constraint this will now fail early and ask him to let us know he was using the this we hope dead code.

PR #311, opened 2020-10-14
ZoneKeyAllocator fixes

Christian-B:
This changed the group key allocation to be ApplicationVertex and Partition Identifiers based.
    This to provide better support when splitter result in multiple partitions associated with an ApplicationVertex,

There is also now support for FixedKeyAndMaskConstraint

In most cases the Pair ApplicationVertex and Partition Identifiers will equate to one Application Level Partition.

The data is however generated from the MachineGraph and using the Machine Level Partition Identifier which allows splitters/partitioners the right to not exactly map Application level Partitions and Machine level ones.
It also allow the addition of internal machine edges/partitions without needing to have matching application ones.

If the machine prevertex in machine edge has no ApplicationVertex, the MachineVertex is used as the "app_id" whcih works but means it can not be grouped. this also support GFE now.

The Global approach to key zones is now done by the same code. It is now more flexible as it can handle the Retina while still leaving as many other vertexes as possible with global zones.

tested by: https://github.com/SpiNNakerManchester/sPyNNaker8/pull/475

First part of the replacment for https://github.com/SpiNNakerManchester/PACMAN/pull/308


pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
 
     .. note::
-        No constraints are supported, and that the number of keys
-        required by each edge must be 2048 or less, and that all edges coming
-        out of a vertex will be given the same key/mask assignment.
+        No special constraints (except ContiguousKeyRangeContraint) are
+        supported.  This will only work if the numbers above add up to 32-bits
+        (an error will result if not).  A single large vertex (like a retina)
+        and lots of small vertices (like a large neural network)
+        will thus not likely work here.
```
rowleya, line 73: I think you have fixed this...

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
-                    app_vertex.machine_vertices))
-                max_app_keys_bits = max(
-                    max_app_keys_bits, machine_bits + key_bits)
-                key_bits_per_app[app_vertex] = key_bits
-        source_bits = self.__bits_needed(source_zones)
-
-        if source_bits + max_app_keys_bits > BITS_IN_KEY:
+        for app_id in progress.over(by_app_and_partition_name):
+            by_app = by_app_and_partition_name[app_id]
+            for partition_name, by_partition_name in by_app.items():
```
rowleya, line 181: "by_partition_name" tripped me up a bit... might be better to simply be "machine_vertices"

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
+                self.__fixed_keys.append((key, n_keys))
+
+    def __calculate_zones(self):
         """
-        :return: tuple containing max app keys and a map of app to max keys
-            for any given machine vertex
-        :rtype: tuple(int, dict(ApplicationVertex, int))
         :raises PacmanRouteInfoAllocationException:
         """
+        by_app_and_partition_name = \
```
rowleya, line 173: Is that simpler if it is just "machine_partitions"?

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
+                    len(self.__fixed_used))
+                if app_part_bits + self.__n_bits_atoms_and_mac <= BITS_IN_KEY:
+                    return  # working result found
+            self.__n_bits_atoms -= 1
+
+        raise PacmanRouteInfoAllocationException(
+            "Unable to use ZonedRoutingInfoAllocator please select a "
+            "different allocator")
+
+    def __check_fixed(self):
```
rowleya, line 257: The side effect of this method might be more obvious if it returned the set it created and then the assignment could be done outside.  It might then be called something like "get_fixed_used" as it currently doesn't do any checking.

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
-        max_app_keys_bits, key_bits_per_app = self._calculate_zones()
+        self.__find_fixed()
+        self.__calculate_zones()
+        self.__check_zones()
+        return self.__allocate()
 
-        return self._simple_allocate(max_app_keys_bits, key_bits_per_app)
+    def __find_fixed(self):
+        by_app_and_partition_name = \
+            self.__machine_graph.multicast_partitions
```
rowleya, line 149: As mentioned elsewhere, would this be better as simply "machine_partitions"?

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
+        for app_id in by_app_and_partition_name:
+            by_app = by_app_and_partition_name[app_id]
+            for partition_name, by_partition_name in by_app.items():
```
rowleya, line 152: As mentioned elsewhere, would "by_partition_name" be better as simply machine_vertices?

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
+                max_keys = 0
+                for mac_vertex in by_partition_name:
+                    partition = self.__machine_graph.\
+                        get_outgoing_edge_partition_starting_at_vertex(
+                            mac_vertex, partition_name)
+                    n_keys = self.__n_keys_map.n_keys_for_partition(partition)
+                    max_keys = max(max_keys, n_keys)
+                if max_keys > 0:
```
rowleya, line 189: Does this also need to filter out the fixed_keys? As otherwise they will get in the way of the calculations?

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
+                    return  # working result found
+            self.__n_bits_atoms -= 1
+
+        raise PacmanRouteInfoAllocationException(
+            "Unable to use ZonedRoutingInfoAllocator please select a "
+            "different allocator")
+
+    def __check_fixed(self):
+        self.__fixed_used = set()
+        bucket_size = 2 ** self.__n_bits_atoms_and_mac
```
rowleya, line 259: I think if this is "global", the space would be slightly different, as the maximum of (n_atoms + n_mac) is not the same as the maximum of n_atoms + the maximum of n_mac.

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
+                    app_part_bits, self.__n_bits_atoms_and_mac))
+
+        if not self.__flexible:
+            if app_part_bits + self.__n_bits_machine + self.__n_bits_atoms > \
+                    BITS_IN_KEY:
+                # We know from above test that all will fit if flexible
+                # Reduce the suggested size of n_bits_atoms
+                self.__n_bits_atoms = \
+                    BITS_IN_KEY - app_part_bits - self.__n_bits_machine
+                self.__n_bits_atoms_and_mac = BITS_IN_KEY - app_part_bits
```
rowleya, line 226: I think this just means that it doesn't fit... how can you reduce the number of bits used by the atoms if we need to keep the mask the same?

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
+            else:
+                # Set the size of atoms and machine for biggest of each
+                self.__n_bits_atoms_and_mac = \
+                    self.__n_bits_machine + self.__n_bits_atoms
+
+    def check_flexible_zones(self):
```
rowleya, line 232: Is this ever actually called?

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
-                        routing_infos.add_partition_info(
-                            PartitionRoutingInfo([key_and_mask], partition))
-                by_app_vertex[app_vertex] = BaseKeyAndMask(
-                    base_key=app_key, mask=app_mask)
-
-        return routing_infos, by_app_vertex
+        app_part_index = 0
+        for app_id in progress.over(by_app_and_partition_name):
+            while app_id in self.__fixed_used:
+                app_part_index += 1
```
rowleya, line 282: I could be wrong, but I think this loop, once entered, will never end

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
+                    n_bits_atoms = self.__atom_bits_per_app_part[
+                        (app_id, partition_name)]
+                    if n_bits_atoms < self.__n_bits_atoms:
+                        # Ok it fits use global sizes
+                        n_bits_atoms = self.__n_bits_atoms
+                        n_bits_machine = self.__n_bits_machine
+                    else:
+                        # Nope need more use that so adjust n_bits_machine down
+                        n_bits_machine = \
+                            self.__n_bits_atoms_and_mac - n_bits_atoms
```
rowleya, line 301: As above, I'm not sure how this makes it global

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
-                            PartitionRoutingInfo([key_and_mask], partition))
-                by_app_vertex[app_vertex] = BaseKeyAndMask(
-                    base_key=app_key, mask=app_mask)
-
-        return routing_infos, by_app_vertex
+        app_part_index = 0
+        for app_id in progress.over(by_app_and_partition_name):
+            while app_id in self.__fixed_used:
+                app_part_index += 1
+            for partition_name, by_partition_name in \
```
rowleya, line 283: See previous comment re: naming

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
-        :param dict(ApplicationVertex,int) key_bits_map:
-            map of app vertex to max keys for that vertex
+        :param dict((ApplicationVertex, str),int) mask_bits_map:
+            map of app vertex,name to max keys for that vertex
         :return: tuple of routing infos and map from app vertex and key masks
-        :rtype: tuple(RoutingInfo, dict(ApplicationVertex,BaseKeyAndMask))
+        :rtype: tuple(RoutingInfo,
+            dict((ApplicationVertex, str), BaseKeyAndMask))
         """
+        by_app_and_partition_name = \
```
rowleya, line 274: See previous comment re: naming

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,308 @@
...
+                    else:
+                        # Nope need more use that so adjust n_bits_machine down
+                        n_bits_machine = \
+                            self.__n_bits_atoms_and_mac - n_bits_atoms
+
+                for machine_index, vertex in enumerate(machine_vertices):
+                    partition = self.__machine_graph.\
+                        get_outgoing_edge_partition_starting_at_vertex(
+                            vertex, partition_name)
+                    if partition in self.__fixed_partitions:
```
rowleya, line 307: The above code does quite a lot of stuff which is then ignored if this is true...

rowleya (changes requested):
A couple of things spotted I think - I might be confused about them of course!

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: agreed. Fixed

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: I added a comment first tie it is used.

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: renamed to set_fixed_used

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: There are two levels of map
Key level 1 app_id
key Level 2 partitoin_name
Value  list of vertices

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: Good spot.

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
rowleya: The issue is more that the name doesn't say *what* is by app and partition name.... just that something is!  You have to go back to this line to know what it is... Maybe then m_partition_map or something...

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: Choose not to rename due to there being TWO levels.

Happy to rename if two names are suggested.

rowleya: I am happy with "partition_name", just "by_partition_name", which at this point is hard to trace... it is machine_vertices I believe; they are not "by_partition_name" at this point because you already have the partition_name here.

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
rowleya: If you want to go all Alan here, you could call it machine_partitions_by_app_and_partition_name... I don't really recommend this though!

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: Yes that is true. 
The issue is the above code is done once for each list of vertices and the constraints are at the single vertex / partition level.

This was a case of lets keep the code simple even if in the rare case of a fixed it has done a bit of work it throws away.
it does avoid a check higher up.

Christian-B:
Did the dict renaming after all.
 multicast_partitions is a map of app_id to paritition_vertices
 paritition_vertices is a map of partition(name) to set(vertex)

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: name changed

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: rename attempted

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: commented added to say in this case it is not global!

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: renamed

rowleya:
There seem to be some unresolved comments here (github kindly hid them for you).

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
rowleya: This is still not fixed in the version I am looking at...

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
rowleya: Looking again at this, I can't help but think it will go wrong then you get one of the vertices whose atoms happen to be bigger than can be represented by n_bits_atoms...

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,309 @@
...
+            return  # felibible same as none flexible
+
+        raw_app_part_bits = self.__bits_needed(
+            len(self.__atom_bits_per_app_part))
+
+        while self.__n_bits_atoms >= 0:
+            self.__n_bits_atoms_and_mac = \
+                self.__n_bits_machine + self.__n_bits_atoms
+            if self.__n_bits_atoms_and_mac + raw_app_part_bits <= BITS_IN_KEY:
+                self.__set_fixed_used()
```
rowleya, line 251: Now I see this is being used, a couple of extra comments.  Firstly, this seems to be called twice.  I can't see it doing anything different the second time, but I could be wrong.

Christian-B: The second call was well OOPS!  Good spot thanks!

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -16,154 +16,303 @@
...
+                machine_vertices = list(paritition_vertices)
+                machine_vertices.sort(key=lambda x: x.vertex_slice.lo_atom)
+                if self.__flexible:
+                    n_bits_atoms = self.__atom_bits_per_app_part[
+                        (app_id, partition_name)]
+                    n_bits_machine = self.__n_bits_atoms_and_mac - n_bits_atoms
+                else:
+                    n_bits_atoms = self.__atom_bits_per_app_part[
+                        (app_id, partition_name)]
+                    if n_bits_atoms < self.__n_bits_atoms:
```
rowleya, line 288: Can probably be equal too?

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
rowleya: OK, I can see how this could help; if you can't fit the global, you make it slightly more flexible.  It might help, but it might not...  I guess a warning could be useful in this case?

rowleya (changes requested):
Almost, just a couple of comments

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
Christian-B: I think the result is the same. But yes it reads better with the equal here. Fixed.

Christian-B:
If global does not fit going to semi flexible is never worse as the alternative is to give up.
Logger warning added as requested,

rowleya (approved):

PR #318, opened 2020-10-28
expose application_level_used

Christian-B:
This avoids algorithms having to be passed in an application graph just to check if they should also use the application level


rowleya:
Nothing wrong with having this, but then you can actually just ask if the elements of the machine graph have application equivalents.  If we actually need an application graph and not just the elements of it, there is already a problem...

Christian-B:
No longer needed

PR #319, opened 2020-10-28
order_vertices_for_dependent_splitters fix and test

Christian-B:


alan-stokes:
this is missing the test where the splitter is set by the selector not the direct vertex

alan-stokes:
ignore me. was mentally at a differnet place

PR #324, opened 2020-11-05
turn on python actions

Christian-B:
Includes https://github.com/SpiNNakerManchester/PACMAN/pull/325

So when this goes in close that too


andrewgait (approved):
Looks good to me, merging...

PR #326, opened 2020-11-12
Sdram partition merge

alan-stokes:
used for help


Christian-B:
I would do it the other way arround from cherry into sdram

Christian-B:
no longer needed cherry is master

PR #337, opened 2020-12-04
Version binding for numpy

dkfellows:
From https://numpy.org/devdocs/release/1.20.0-notes.html

> support for Python 3.6 has been dropped

The real details are in https://numpy.org/neps/nep-0029-deprecation_policy.html


rowleya:
Minor point:  if you want to allow the version to evolve for later Python versions, you can do this:
https://stackoverflow.com/questions/19559247/requirements-txt-depending-on-python-version

I don't actually know if we do want to allow this though...

dkfellows:
I was already working on that. I knew I'd already done it somewhere… but I had to look up where. 😉

doc/source/conf.py
```diff
@@ -56,7 +56,7 @@
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3.6', None),
-    'numpy': ("https://numpy.org/doc/stable/", None),
+    'numpy': ("https://numpy.org/doc/1.19/", None),
```
Christian-B, line 59: I would rather have seen us moving sphinx to a later python version but that is out fo scope of this PR

Christian-B (approved):
Merge when green!

doc/source/conf.py
dkfellows: That controls what version of the numpy documentation we point to in our types.

Christian-B:
If this works, As a second PR see is adding a version for 2.7 gets setup to work without the need to install requirement. But 1, AFTER this PR and 2 only if easy

dkfellows:
Feel free to write the PR. The version range binding trickery I've used here is accepted in both `requirements.txt` and `setup.py`; it's totally Python standard and has been for ages.

Christian-B:
> Feel free to write the PR. The version range binding trickery I've used here is accepted in both `requirements.txt` and `setup.py`; it's totally Python standard and has been for ages.

A quick check:
https://github.com/SpiNNakerManchester/PACMAN/runs/1509289122?check_suite_focus=true

Shows there is a lot more than just numpy broken in python 2,7 so not worth it.

PR #343, opened 2021-01-05
Repair SDRAM Allocation

rowleya:
This fixes SDRAM allocation by making the splitter do the allocation rather than trying to work this out automagically.  At the splitter, this is actually really easy!


Christian-B:
Best to check with @alan-stokes As this apears to remove the tracking by chip he believed he needed for the sdram stuff

rowleya:
@alan-stokes did request that we look at the Poisson-over-SDRAM integration first.  From doing most of the work on this (not quite ready), I can see that the code that I have deleted here would not have helped at all.  The issue is that you can't just shove the SDRAM into the resource tracker and hope to get it right.  You have to instead add all the SDRAM at once to get the correct allocation.  There isn't an easy way to do that from inside, but from outside it is much easier.

rowleya:
Note, I also snuck a change to the error message when resources are exhausted here; this displays the remaining resources board-by-board and chip-by-chip as otherwise I was getting a situation where I needed to allocate 3 cores and 3 were available, but not on the same chip; the message wasn't clear that this was the case.

alan-stokes:
meh. can be accepted

PR #346, opened 2021-01-18
sortedcollections needs version restriction for Python 2.7

dkfellows:
This is blocking downstream builds, and is caused by sortedcollections 2 (strictly 2.0.1 as 2.0.0 wasn't released) dropping Python 2.7 support. Which wasn't loudly announced.


Christian-B:
better fix than https://github.com/SpiNNakerManchester/PACMAN/pull/345

PR #356, opened 2021-02-17
Lost spike issues

rowleya:
Adds an incoming route, allowing the detection of default routes


andrewgait (approved):
Looks fine to me, good to have a test as well.

andrewgait:
Needs merging with master before it can be merged.

PR #361, opened 2021-03-04
ask machine for unused xy

Christian-B:
Fixes https://github.com/SpiNNakerManchester/PACMAN/issues/360

Depends on: https://github.com/SpiNNakerManchester/SpiNNMachine/pull/156
There is a second branch on SpiNNMachine with a name matching the branch in this pr


rowleya (approved):
This is fine, but leaves behind a lot of unused code!

Christian-B:
> This is fine, but leaves behind a lot of unused code!

Sorry don't know what code is now unused.

self._free_space_tracker is used in ElementAllocatorAlgorithm._find_slot
which is called by
CompressibleMallocBasedRoutingInfoAllocator._allocate_keys_and_masks

self._allocate_element is used by self.allocate_chip_ids which is used by call

rowleya:
This originally made use of the ElementAllocatorAlgorithm code; before these changes it would eventually call self._allocate_elements which would do the actual allocation from a pool of elements stored in the superclass.  To facilitate this, the init would say that there are 2**32 elements available, which would then create the pool.  During call, the code then goes through the machine and allocates the existing chips from this pool.  It then finally goes through the graph and calls the other methods to do the allocation.

To simplify:
 1. The class no longer needs to extend ElementAllocatorAlgorithm at all, and probably therefore needs no init at all.
 2. The call method can avoid going through the machine allocating the chips therein.

So the changes are not huge, but should save quite a bit of time, especially on a big machine.

Christian-B:
code simplified as requested

rowleya:
I think the code simplifications were not pushed... I have done these though and it still seems to work as expected

PR #370, opened 2021-04-19
fixed call output docs

Christian-B:


dkfellows (approved):
If that is the correct type (I tend to get lost in there) then this is good.

PR #379, opened 2021-05-11
Makes Sphinx more like other repo

Christian-B:
Inlcudes fixes from https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/806

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -28,7 +28,8 @@
 # serve to show the default.
 
 import os
-import inspect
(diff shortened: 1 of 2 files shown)


dkfellows:
That looks fine at first glance. I _think_ the code that got replaced was the remnants of stuff that was there for debugging what's going on inside Sphinx…

The proof of the pudding is in the eating. Do the docs look right? If so, we're fine! 😆

PR #385, opened 2021-06-01
Exceptions cleanup

Christian-B:
Looking at coverage report leads to some Exceptions cleanup.

PacmanProcessorNotAvailableError never used!
PacmanAlgorithmFailedToCompleteException never used 
    ASB checks for it so that will be cancelled in pr
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/820

MinimisationFailedError as sometimes called with the expected param but more often with a String msg.
Changed to be just a String msg as most other Exceptions.

Tested by 
https://github.com/SpiNNakerManchester/IntegrationTests/pull/59


rowleya (approved):
Looks like the functionality here has been replaced with a logger.error, so yes, remove from here.

PR #396, opened 2021-06-18
Split synapse neuron models

rowleya:
A couple of extra things needed for this to pass tests correctly


andrewgait (approved):
Looks fine to me.

PR #398, opened 2021-06-24
Speedup Big FromListConnector 

Christian-B:
Needed for https://github.com/SpiNNakerManchester/sPyNNaker/pull/1095

Application_vertex caching slices has two benefits
1. Avoid recreating then every time
2. FromListConnector  slices == slices is the id based

could_connect change to vertexes
1. Allows FromList to use __split_conn_list which is MUCH faster!


andrewgait (approved):
Looks good to me.

PR #400, opened 2021-07-01
Pylint fix

Christian-B:
pylint picked up that label could be none


Christian-B:
minor fix and master broken so self merging

PR #403, opened 2021-07-06
Stop ResourceTracker from looking at the Whole Machine

Christian-B:
The ResourceTracker had been designed to consider a many pre allocations on a few Chips.

In Practice only cores, sdram and iptags where ever reserved.
Also the reservation was the same for all chips with a bit extra for all ethernet chips.

So this PR.
1. Remove the unused preallocation types
2. Changes Preallocation to be just two types all and ethernet
3. The ResourceTracker now only tracker the Boards/Chip it needs to,
    a. if given a list of Chips it tracks all
    b. It not just tracks when it sees an x,y or ip/board_address

One minor additional lost of functionality is that if the SlicePartitioner allocates to a Chip and then deallocates that chip resources are freed but the chip will always be counted as used. In theory this could result in an extra board being requested from spalloc

testing with spalloc and VA_benchmark shows the difference.
PreAllocateResourcesForExtraMonitorSupport
from  0:00:00.281158 to 0:00:00.000409
 SpYNNakerSplitterPartitioner
from 0:00:04.748884 to 0:00:00.017941

As much of the saved time is down to creating a lot less objects there is a memory gain too


dkfellows (approved):
Looks good to me

Christian-B:
Fixes https://github.com/SpiNNakerManchester/PACMAN/issues/394

PR #404, opened 2021-07-14
Algorithms cleanup

Christian-B:
Part of https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/834


Christian-B:
This is outdtaed due to the no executor work. But please leave the branches so that the idea can be merged in there

PR #405, opened 2021-07-14
If an optional param/token can be provided it must be

Christian-B:
I discovered a toekn optionally required loop

ChipIOBufClearer had an input <token>ReadIOBuf, ChipIOBufExtractor had an input <token>ApplicationRun and an has an input <token>ClearedIOBuf

This resulted in PACMANAlgorithmExecutor picking the order by implementation rather than by design.

This was caused by the support for an algorithm that inputs and outputs the same ParamType

Removal of this safety requires https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/835

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/68


Christian-B:
This is no longer needed as no executor will nuke tokens

PR #411, opened 2021-10-27
No need for code in case there is no app graph

Christian-B:
Part of https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/884

adds two app graph functions:
machine_vertices and n_machine_vertices
to make it easier to access these from the app graph

removes code in case there is no app graph


Christian-B:
done elsewhere

PR #424, opened 2022-03-17
No need for "Ordered" dicts

Christian-B:
As of python 3.7 (the first version we officially support) all dicts are now order guaranteed.
Even python 3.6 had it just not made if a guarantee

The OrderedDict and our DefaultOrderedDict have a few minor advantages
such as ability to iterate reversed. (python 3.8 normal dicts can but not 3.7 or 3.6)
Ability to .move_to_end() In normal dicts you have to remove and readded to get that affects
But are about 30% slower and do use more ram
ref: https://realpython.com/python-ordereddict/

This set of PRs changes the "ordered" to normal as far as  possible.

They are all independent except that this PACMAN one must be done before (or at the same time as) the utils one!

OrderDict was kept in SpinUtils Ordered Sets as that supports reverse.  

SpallocServer not done as 1. it uses move_to_end and 2 is actively being replaced with a Java version.

tested by:

TODO!

### DIFF
--- a/pacman/model/graphs/abstract_multiple_partition.py
+++ b/pacman/model/graphs/abstract_multiple_partition.py
@@ -13,8 +13,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from collections import OrderedDict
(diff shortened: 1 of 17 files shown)


Christian-B:
wrong branch name so can not test other prs

PR #431, opened 2022-05-12
merge graph into ApplicationGraph

Christian-B:
Now that there is no MachineGraph no need for a common Graph object.

Decided to keep the class name ApplicationGraph to show it can only take Application Objects

Note: Also removed ConstrainedObject as a super class as we can not even remember why or how a Graph could be Constrained.


dkfellows:
> Also removed ConstrainedObject as a super class as we can not even remember why or how a Graph could be Constrained.

👍 👍 👍

Christian-B:
Better solution found

PR #432, opened 2022-05-17
Ro one graph2

Christian-B:
Builds on https://github.com/SpiNNakerManchester/PACMAN/pull/431

The graph itself wraps the machine vertex and machine edges

If approved the wraps in GFE can be removed.


Christian-B:
better solution found

PR #433, opened 2022-05-26
app vertex max_atoms_per_core

Christian-B:
every app vertex had one or even move MaxVertexAtomsConstraint
no one used  FixedVertexAtomsConstraint

So simplify and just make it a property on app vertex.

---

One change we now allow the raising of the max_atom_per_core

https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/945
https://github.com/SpiNNakerManchester/sPyNNaker/pull/1184


pacman/model/graphs/application/application_vertex.py
```diff
@@ -184,13 +184,21 @@ def vertex_slices(self):
...
-                return constraint.size
+        return self._max_atoms_per_core
+
+    def set_max_atoms_per_core(self, new_value):
+        """
+        Set the max_atoms_per_core.
+
+        Can be used to raise the number of atoms.
+
+        :param int new_value: Value to set if not higher than previous
```
rowleya, line 199: Wrong description - it can be higher or lower

pacman/model/partitioner_splitters/abstract_splitters/abstract_splitter_common.py
```diff
@@ -66,7 +60,6 @@ def __init__(self, splitter_name=None):
         if splitter_name is None:
             splitter_name = self.DEFAULT_SPLITTER_NAME
         self._splitter_name = splitter_name
-        self._max_atoms_per_core = sys.maxsize
         self._is_fixed_atoms_per_core = False
```
rowleya, line 63: Could also get rid of this maybe?

rowleya (approved):
Looks fine - some minor changes suggested but not critical

pacman/model/partitioner_splitters/abstract_splitters/abstract_splitter_common.py
rowleya: (was meant to point at self._is_fixed_atoms_per_core)

Christian-B: Yes! but!
yes you are correct.
But also need to strip it out of AbstractSplitterSlice

pacman/model/graphs/application/application_vertex.py
Christian-B: agreed fixed

PR #450, opened 2022-07-26
Create graph

Christian-B:
Builds on https://github.com/SpiNNakerManchester/PACMAN/pull/449

An ApplicationGraph is now created during the Writer.setup or mock call. So except for before sim.setup (or unittest_setup())
there is always a graph so no need to call create graph or set_(runtime)_ graph

The Graph Class (or at least what was actually needed) has been merged into ApplicationGraph.
With the following changes
- The Graph no longer has a label so all graph label params removed
- You can no longer set or get constraints on graph (Was totally unused)
- get_edges_ending_at_vertex method removed (no known use)
- add_vertices and add_edges methods removed as only used in tests
- Vertex are now only held as the Values in a dict.
    (This has a minor cost that you can not iterate vertices and add at the same time)
- The allowed Vertex and Edge type to add are now harded coded.

As the Graph is never cloned the clone method has been removed.

As the Graph is never given to the use the ApplicationGraphView is no longer need

Must be done after:
https://github.com/SpiNNakerManchester/SpiNNUtils/pull/184

https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/961
https://github.com/SpiNNakerManchester/sPyNNaker/pull/1202
https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/217

tested by
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/create_graph/3/pipeline

Documented by
https://github.com/SpiNNakerManchester/SpiNNakerManchester.github.io/pull/42


pacman/data/pacman_data_view.py
```diff
@@ -219,15 +210,12 @@ def get_n_vertices(cls):
...
-                cls.__pacman_data._graph = ApplicationGraph("Mocked")
-            else:
-                raise cls._exception("graph")
+            raise cls._exception("graph")
         return cls.__pacman_data._graph.n_vertices
 
     @classmethod
     def iterate_partitions(cls):
-        """ The partitions in the user application graph.
+        """ The partitions in the user application graphs an iterator
```
andrewgait, line 218: I think you mean "graph as an iterator"?

pacman/data/pacman_data_view.py
```diff
@@ -271,69 +253,42 @@ def get_outgoing_edge_partitions_starting_at_vertex(cls, vertex):
...
         return sum(len(vertex.machine_vertices)
-                   for vertex in cls.get_runtime_graph().vertices)
+                   for vertex in cls.__pacman_data._graph.vertices)
 
     @classmethod
-    def get_runtime_machine_vertices(cls):
+    def iterate_machine_vertices(cls):
         """
-        Gets the Machine vertioces viua the application graph
+        Iterate the Machine vertices viua the application graph
```
andrewgait, line 285: "via"

pacman/model/graphs/application/application_graph.py
```diff
@@ -18,44 +18,155 @@
...
+        :param ApplicationVertex vertex: The vertex to add
+        :raises PacmanInvalidParameterException:
+            If the vertex is not of a valid type
+        :raises PacmanConfigurationException:
+            If there is an attempt to add the same vertex more than once
+        """
+        if not isinstance(vertex, ApplicationVertex):
+            raise PacmanInvalidParameterException(
+                "vertex", str(vertex.__class__),
+                "Only ApplicationVertexs can be added")
```
andrewgait, line 58: "Only an ApplicationVertex"

pacman/model/graphs/application/application_graph.py
```diff
@@ -18,44 +18,155 @@
...
+                raise PacmanAlreadyExistsException("vertex", vertex.label)
+            vertex.set_label(vertex.label + self._label_postfix())
+        vertex.addedToGraph()
+        self._vertex_by_label[vertex.label] = vertex
+
+    @property
+    def vertices(self):
+        """ The vertices in the graph.
+
+        :rtype: iterable(AbstractVertex)
```
andrewgait, line 73: Perhaps say that this is a dict rather than just an iterable?

andrewgait (approved):
This looks fine to me, just a few minor edits suggested

pacman/model/graphs/application/application_graph.py
Christian-B: The call returns only the values(0 of the dict not the whole dict. So iterable is correct

PR #452, opened 2022-08-02
remove unused resources

Christian-B:
This PR and associated removes the dtcm and cpu cost models,
As well as the resource containers that held them.
Iptags where moved.

The models where completely inaccurate with guesses, blindly copied guesses and zeros.

With the move to fixed size partitioning the only place that pretended to use these costs has been removed.

The only remaining place that uses cost models (determining auto pause loop lengths) require very accurate or at least never too low models like we have with sdram

Having done a PhD on cost models for a limited part of much simpler sensor networks I can say with certainty that creating dtcm and especially cpu models with a level of accuracy need to make them useful is completely outside of the scope of the HBP project. They are both orders of magnitudes harder than sdram costs.

---
The classes CPUCyclesPerTickResource, DTCMResource and ResourceContainer have been deleted.

Methods that used to return a ResourceContainer have been renamed and now return an AbstractSDRAM implementation
resources_required -> sdram_required
get_resources_used_by_atoms -> get_sdram_used_by_atoms

Methods that used to have a resource_required param now have an sdram param
create_machine_vertex
chip_counter.add_core
init methods of several classes that stored cost models

New properties have been added to MachineVertex
iptags
reverse_iptags

They default to empty lists so should be overridden when there are iptags.

https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/964
https://github.com/SpiNNakerManchester/sPyNNaker/pull/1205
https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/218
https://github.com/SpiNNakerManchester/MarkovChainMonteCarlo/pull/32
https://github.com/SpiNNakerManchester/SpiNNaker_PDP2/pull/62
https://github.com/SpiNNakerManchester/SpiNNGym/pull/48
https://github.com/SpiNNakerManchester/sPyNNaker8NewModelTemplate/pull/87

tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/124


pacman/model/graphs/machine/machine_spinnaker_link_vertex.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2019 The University of Manchester
+# Copyright (c) 2017-2012 The University of Manchester
```
andrewgait, line 1: 2022 surely? ;)

pacman/operations/tag_allocator_algorithms/basic_tag_allocator.py
```diff
@@ -58,18 +58,20 @@ def basic_tag_allocator():
...
-        for iptag in resources.iptags:
+        for iptag in placement.vertex.iptags:
             alloc_chip, tag = __get_chip_and_tag(
                 iptag, eth_chip, tags_on_chip, machine, tags_available)
             tags.add_ip_tag(
                 __create_tag(alloc_chip, placement, iptag, tag),
                 placement.vertex)
-        for reverse_iptag in resources.reverse_iptags:
+        a = placement.vertex.reverse_iptags
+        if a is None:
```
andrewgait, line 72: Is this left over from debugging?

andrewgait (changes requested):
Looks fine apart from a print statement left in by mistake

PR #455, opened 2022-08-04
Constraint simplification.

Christian-B:
https://github.com/SpiNNakerManchester/PACMAN/pull/463 removes the 
FixedKeyAndMaskConstraint

Which means there will only be 1 type of constraint of which there can be at most one.
Farther simplification is possible before merging this!

This PR simplifies the Constraint stack.

There where only two Constraints still being used.
FixedKeyAndMaskConstraint
ChipAndCoreConstraint

Both could probably be done in a better way than Constraints but that is out of scope for this PR.

This PR removes a level of Abstract and removes or simplifies some of the checks on Constraints.

Movie the two remaining Constraint to pacman/model/constraints

Should be done at the same time as:
https://github.com/SpiNNakerManchester/sPyNNaker/pull/1207
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/966
https://github.com/SpiNNakerManchester/SpiNNakerGraphFrontEnd/pull/220
https://github.com/SpiNNakerManchester/MarkovChainMonteCarlo/pull/33

tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/126

fixes https://github.com/SpiNNakerManchester/PACMAN/issues/440
fixes https://github.com/SpiNNakerManchester/PACMAN/issues/439


pacman/utilities/utility_calls.py
```diff
@@ -63,45 +49,6 @@ def _is_constraint_supported(constraint, supported_constraints):
                for supported_constraint in supported_constraints)
 
 
-def check_algorithm_can_support_constraints(
```
andrewgait, line 66: I think if this is being deleted then _is_constraint_supported (above it) can go as well.

pacman/utilities/utility_calls.py
```diff
@@ -15,21 +15,7 @@
 
 import hashlib
 import numpy
-from pacman.exceptions import (
-    PacmanInvalidParameterException, PacmanValueError)
-
-
-def locate_constraints_of_type(constraints, constraint_type):
```
andrewgait, line 22: It doesn't look like locate_first_constraint_of_type(..) (the next function in this file) is used anywhere either so this can probably be deleted.

andrewgait (changes requested):
Looks fine, just wondering whether more can be deleted.

Christian-B:
All Constraints to be removed.

PR #458, opened 2022-08-21
removed n_atoms from LegacyPartitionerAPI

Christian-B:
Fixes https://github.com/SpiNNakerManchester/PACMAN/issues/457

The  LegacyPartitionerAPI defined the property n_atoms which the unnecessary as ApplicationVertex does too and an LegacyPartitionerAPI must be an ApplictionVertex for it to work.

This PR removes the property from   LegacyPartitionerAPI as well as the the auto detection of a Vertex which could take a Legacy Splitter as that was never used.

https://github.com/SpiNNakerManchester/sPyNNaker/pull/1212

Also flagged up:
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/972

Tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/130


Christian-B:
This now only a minor fix (optional) as we have decided to strip out the autodetection of Legacy Vertices that the declaration of n_atoms in  LegacyPartitionerAPI made impossible to do correctly.

PR #461, opened 2022-08-30
Splitter name

Christian-B:
Fixes https://github.com/SpiNNakerManchester/PACMAN/issues/459

Removes spltter name. and several string constants which where just the str(type(self))

Instead AbstractSplitterCommon now does str and repr based on the splitter type

https://github.com/SpiNNakerManchester/sPyNNaker/pull/1216

tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/132


pacman/model/partitioner_splitters/splitter_one_to_one_legacy.py
```diff
@@ -45,23 +45,17 @@ class SplitterOneToOneLegacy(AbstractSplitterCommon):
...
-    def __repr__(self):
-        return self.__str__()
-
     @overrides(AbstractSplitterCommon.set_governed_app_vertex)
     def set_governed_app_vertex(self, app_vertex):
         super().set_governed_app_vertex(app_vertex)
         self._vertex_slice = Slice(0, self._governed_app_vertex.n_atoms - 1)
         self._sdram = (
-            self._governed_app_vertex.get_sdram_used_by_atoms(
+             self._governed_app_vertex.get_sdram_used_by_atoms(
```
andrewgait, line 58: github is saying an extra space has been added here which seems to mess up the formatting...

pacman/model/partitioner_splitters/abstract_splitters/abstract_splitter_common.py
```diff
@@ -39,21 +35,18 @@ class AbstractSplitterCommon(object, metaclass=AbstractBase):
...
 
     def __str__(self):
-        return (
-            f"{self._splitter_name} governing app vertex"
-            f" {self._governed_app_vertex}")
+        try:
+            return (
+                f"{type(self).__name__} on {self._governed_app_vertex.label}")
+        except AttributeError:
+            return f"{type(self).__name__} no vertex"
```
andrewgait, line 49: I would probably be more explicit here - "has no governed_app_vertex" or something like that.

andrewgait (changes requested):
Looks fine, just a couple of minor points.

pacman/model/partitioner_splitters/splitter_one_to_one_legacy.py
Christian-B: reversed

andrewgait (approved):

PR #470, opened 2022-09-23
Fix routing

rowleya:
Fixes the routing to work:
1) When there is an edge between two vertices which share more than one chip.  This can happen with FPGA inputs since the input can come from multiple FPGA links.
2) When there is an edge to a vertex which in reality only targets one machine vertex.  This allows for application vertices that have been placed on multiple non-connected chips, but where source application vertices only target one of those machine vertices (example is LivePacketGather in FEC, once the other PRs have been merged).

Note this also refactored the application router code into several function calls so the changes below may appear larger than they are.

See also:
SpiNNakerManchester/SpiNNFrontEndCommon#978
SpiNNakerManchester/sPyNNaker#1222


Christian-B (approved):
As far as I see them the changes look good.

PR #471, opened 2022-09-26
get_placement_on_processor

Christian-B:
The view.get_placement_on_processor  was missing and needed by ASB.__recover_from _error 

As view.get_vertex_on_processor differed only by where the .vertex was placed I choose to remove the get_vertex_on_processor from both Placements and View and instead replaced its only use with
view.get_placement_on_processor().vertex

Needed for: https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/979


andrewgait (approved):
Looks fine to me

PR #476, opened 2022-10-17
Mvertex label

Christian-B:
The default label of a Machine vertex is the Application vertex label and the slice.

However the slice str has gotten more complex so the labels looks silly.

This pr.
1. simplifies the str of the slice
2. removes the extra test like "MachineVertex for" from the labels.


andrewgait (approved):
This looks like a good idea to me.

PR #478, opened 2022-11-04
Reduce overhead data spec

rowleya:
Fix to ensure that the number of atoms is correctly reported when a fixed key is used.

Note this is part of reduce_overhead_data_spec because it is needed to make that branch work, but it can be merged before as it is independent.


andrewgait (approved):
This looks fine to me

PR #480, opened 2022-11-16
Reduce overhead data spec colours

rowleya:
Adds ability to shift keys, which is useful with coloured spikes


pacman/utilities/utility_calls.py
```diff
@@ -173,19 +177,26 @@ def get_field_based_keys(key, vertex_slice):
     # get the key
     keys = numpy.sum(numpy.left_shift(coords, shifts), axis=1)
 
+    # Do any final shifting as required (zero shift is valid but does nothing)
+    keys = numpy.left_shift(keys, shift)
```
andrewgait, line 181: I don't suppose this call takes very long even if there are lots of keys but it may be worth considering not doing it if the shift is zero.

andrewgait (approved):
Approved with a question / consideration.

pacman/utilities/utility_calls.py
rowleya: I thought it might do nothing in numpy, but it looks like it still does work, so I have added the condition

PR #491, opened 2023-02-21
Fix pool dense

rowleya:
Fixes issue with "overlap" routing where input from a device arrives from multiple chips (i.e. an FPGA) and one of those chips is connected to one subset of machine vertices from an application vertex and a different chip is connected to a different subset of machine vertices!  This can occur if the application vertex has so many machine vertices as to be bisected by the device interface points.


andrewgait (approved):
Looks good to me

PR #496, opened 2023-03-13
A round of technical polishing on the docs

dkfellows:
Little things like full stops at the end of sentences, marking notes properly, making indentation work right, etc. Also converts some use of `String.format()` into f-strings (when it makes things fit on a line more neatly).


dkfellows:
See SpiNNakerManchester/SpiNNUtils#223

PR #499, opened 2023-03-17
A way to ensure that calls into jsonschema only happen from one thread

dkfellows:
A thread pool with a single thread in it to handle the validation. Ugly and full of overhead, but at least it is pretty easy to see that this is doing the right thing. Other threads will simply never touch the package at all.


dkfellows:
#497 and SpiNNakerManchester/SpiNNFrontEndCommon#1032 are related

Christian-B:
appears to no longer be needed.

PR #501, opened 2023-03-23
Set the max atoms per core to something sensible

rowleya:
Fixes the max atoms per core on a SpiNNaker Link input device when the device is split into multiple "cores".  This helps sPyNNaker master population tables get the values right.


andrewgait (approved):
Again this looks fine to me.

PR #505, opened 2023-04-12
Enable spell checking

dkfellows:
See:
* https://github.com/SpiNNakerManchester/SupportScripts/pull/54
* https://github.com/SpiNNakerManchester/SpiNNUtils/pull/225


dkfellows:
That's as many spelling issues as I plan to fix here; fixing the remaining ones is more likely to cover up future problems.

PR #512, opened 2023-04-28
Spif 1d

rowleya:
Fixes to make SPIF output correct


rowleya:
Turns out this isn't needed now (possibly because keys have been corrected)

PR #513, opened 2023-04-28
Type Annotations and Checking

dkfellows:
This PR has checked type annotations for PACMAN, a *substantial* task! The type checking is done with `mypy`, a tool that is as close to "standard" for type checking in Python as any. It forms part of a wider effort to type check the whole stack, which has exposed all sorts of subtle problems along the way.

Also included in this PR are a few bits and pieces of complexity reduction.

* Deletes some unused classes
* Uses the built-in sorting functions
* Avoids the loop-yield pattern
* Renames a confusingly-named class (`Entry`&rarr;`RTEntry`; still not sure if it is needed)
* Builds some strings for comparisons more efficiently

The overall effect seems (on my machine) to make about a 10% difference in the running speed of some tests. These specifically include some very slow-running ones!


Christian-B (changes requested):
The Python Compressors are nearly ONLY used for testing the C code.
(Their use of full back to on host is to the best of my knowledge never used.)

As such I uncomfortable with the use of the python optimisations.

dkfellows:
See also:
* SpiNNakerManchester/SpiNNMan#327

(Depended on by the linked FEC change)

Christian-B:
Done by followup pr https://github.com/SpiNNakerManchester/PACMAN/pull/529

PR #514, opened 2023-05-03
convert 1D shape into largest perfect md shape possible

Christian-B:
Simplest version uses a m D shape with all sides the same size.

So 256 in 3D would be (6,6,6) where a shape of (7,6,6) would be possible


andrewgait (approved):
No objections as long as it doesn't break any current tests.

andrewgait:
Though perhaps a test could also be added using the script at http://spinnakermanchester.github.io/development/2d_convolutions.html

Christian-B:
This is only one of several options.

Another is to keep as master and require the setting of max atoms.

Another is to have 2D defaults

andrewgait:
I don't know how this was resolved so leaving it open and re-assigning it.

Christian-B:
Assume this is no longer needed.

rowleya:
Yes, I think this was a test on the path, but is now no longer needed.

PR #516, opened 2023-05-25
Interdimensional compatibility

rowleya:
Part of SpiNNakerManchester/sPyNNaker#1345.

Fixes multidimensional support to work between vertices of any dimensionality.


Christian-B (approved):
ok as is

PR #518, opened 2023-07-10
Defaults from cfg

Christian-B:
Part of https://github.com/SpiNNakerManchester/SpiNNMachine/pull/215

The ApplicationPlacer uses the cfg Value even once ChipInfo has been read 
This is to decide if it has run out of sdram because the actual machine is too small or because it would be impossible with any machine of this board type.

Highly unlikely with sdram but could happen with n_cores which will use the same pattern.

### DIFF
--- a/pacman/config_setup.py
+++ b/pacman/config_setup.py
@@ -14,29 +14,42 @@
 
 import os
 from spinn_utilities.config_holder import (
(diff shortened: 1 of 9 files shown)


Christian-B:
https://github.com/SpiNNakerManchester/SpiNNMachine/pull/217 was a better method

PR #530, opened 2023-10-30
add code for the muti dimensionall maths

Christian-B:


Christian-B:
do we still need this?

rowleya:
Good question - it was useful to see, but not sure if it is useful now the code is there in a different form...

Christian-B:
Better not to merge as otherwise the code has to be maintained

PR #542, opened 2024-02-20
Allow overlap

rowleya:
I think that recently a change was added to ensure no overlap of fixed keys during allocation.  We actually need to allow this for the sake of the SPIF devices, since commands are sent to multiple devices connected to the same board.  The rule has therefore changed; this is OK so long as they don't overlap at the router level.  This PR should add checks to enforce this.


Christian-B:
I ran a the self synfire with 100,000 neurons in and a giant fromlist connector master and this branch.  Load Application data specification master 3,639,730 vs here. 7,001,228.  This is in Python!

Christian-B (approved):
ok to merge

PR #550, opened 2024-03-14
Remove 18

Christian-B:
Part of https://github.com/SpiNNakerManchester/SpiNNMachine/pull/248

MulticastRoutingTableByPartitionEntry
-incoming processor and link id now stored separately

### DIFF
--- a/pacman/model/routing_table_by_partition/multicast_routing_table_by_partition_entry.py
+++ b/pacman/model/routing_table_by_partition/multicast_routing_table_by_partition_entry.py
@@ -29,7 +29,7 @@ class MulticastRoutingTableByPartitionEntry(BaseMulticastRoutingEntry):
     An entry in a path of a multicast route.
     """
 
-    __slots__ = ["_defaultable", "_incoming"]
+    __slots__ = ["_incoming_processor", "_incoming_link"]
 
     def __init__(self, out_going_links: Union[int, Iterable[int], None],
                  outgoing_processors: Union[int, Iterable[int], None],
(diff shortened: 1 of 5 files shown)


pacman/model/routing_table_by_partition/multicast_routing_table_by_partition_entry.py
```diff
@@ -13,47 +13,33 @@
...
-        # OP = outgoing processors
-        "_links_and_procs", )
+    __slots__ = ["_incoming_processor", "_incoming_link"]
 
     def __init__(self, out_going_links: Union[int, Iterable[int], None],
                  outgoing_processors: Union[int, Iterable[int], None],
                  incoming_processor: Optional[int] = None,
-                 incoming_link: Optional[int] = None):
+                 incoming_link: Optional[int] = None,
+                 spinnaker_route: Optional[int] = None):
```
rowleya, line 37: Do we actually make use of this here? I would have thought it would never be called in this way...

rowleya (approved):

pacman/model/routing_table_by_partition/multicast_routing_table_by_partition_entry.py
rowleya: This might be another case where we could have a "from_processor_and_links" function or something like that.  It might alternatively then hold a SpiNNMachine routing entry rather than extend it, but not sure if that would help things here...

Christian-B: The main need for these is in the Mundy routers.

Allow these to be lost and I think the secondary usage of seeing is a route is defaulable can be handled at init time.

PR #553, opened 2024-03-25
Refactor of the application Placer

Christian-B:
This PR does a cleanup of the Application

Suggest looking at the new code rather than the changes.
https://github.com/SpiNNakerManchester/PACMAN/blob/app_placer3/pacman/operations/placer_algorithms/application_placer.py

The code is intended to be well documented so if not it needs fixing.

No explanation here.

Builds on:
https://github.com/SpiNNakerManchester/PACMAN/pull/552
SpiNNakerManchester/SpiNNFrontEndCommon#1170
SpiNNakerManchester/sPyNNaker#1440

tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/261


rowleya (approved):
I have looked over the algorithm and can see nothing wrong with it from inspection.  It is probably better tested than it was before now anyway!

PR #555, opened 2024-03-28
Single routing entry

Christian-B:
part of https://github.com/SpiNNakerManchester/SpiNNMachine/pull/250


Christian-B:
https://github.com/SpiNNakerManchester/PACMAN/blob/master/unittests/operations_tests/routing_table_generator_tests/test_basic.py need to be checked against the spin2 version once possible

PR #575, opened 2024-10-07
Check partitions

rowleya:
What needs to be done to check all partitions


rowleya:
No longer needed as done elsewhere

PR #576, opened 2024-10-08
Check if Partition d is valid if None

Christian-B:
vertex's that depend on  RoutingInfo.get_first_key_from_pre_vertex to determine if they have send keys will have to Implement  send_partition_ids to avoid a log message

For example PopulationApplicationVertex


Christian-B:
not needed due to https://github.com/SpiNNakerManchester/PACMAN/pull/578

PR #578, opened 2024-10-10
Allow different partitions

rowleya:
Work to make sure partitions that vertices send using, and the partitions actually connected agree!

As the method is deprecated, the other PRs are not strictly necessary, but should avoid deprecation warnings:
 - SpiNNakerManchester/SpiNNakerGraphFrontEnd#286
 - SpiNNakerManchester/SpiNNFrontEndCommon#1219
 - SpiNNakerManchester/sPyNNaker#1496
 - SpiNNakerManchester/SpiNNGym#96
 - SpiNNakerManchester/MarkovChainMonteCarlo#72
 - SpiNNakerManchester/SpiNNaker_PDP2#89
 - SpiNNakerManchester/TSPOnSpiNNaker#48

Tested by:
 - SpiNNakerManchester/IntegrationTests#294

### DIFF
--- a/pacman/model/routing_info/routing_info.py
+++ b/pacman/model/routing_info/routing_info.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 from __future__ import annotations
 from collections import defaultdict
-from typing import Dict, Iterator, Optional, Iterable, TYPE_CHECKING
+from typing import Dict, Iterator, Optional, Iterable, Set, TYPE_CHECKING
 from deprecated import deprecated
 from pacman.exceptions import PacmanAlreadyExistsException
 if TYPE_CHECKING:
@@ -161,6 +161,24 @@ class RoutingInfo(object):
             return False
         info = self._info[vertex]
         return partition_id in info
+
+    def test_pre_vertex_partition_ids(
+            self, vertex: AbstractVertex,
+            allowed_partition_ids: Set[str]):
+        """
+        Check that the partition ids for a vertex are in the allowed set.
+
+        :param AbstractVertex vertex: The vertex to search for
+        :param set[str] allowed_partition_ids: The allowed partition ids
+        :raise KeyError: If the vertex has an unknown partition ID
+        """
+        if vertex not in self._info:
+            return
+        info = self._info[vertex]
+        for partition_id in info:
+            if partition_id not in allowed_partition_ids:
+                raise KeyError(
+                    f"Vertex {vertex} has unknown partition ID {partition_id}")
 
     def get_single_routing_info_from_pre_vertex(
             self, vertex: AbstractVertex) -> Optional[VertexRoutingInfo]:
--- a/pacman/model/routing_info/routing_info.py
+++ b/pacman/model/routing_info/routing_info.py
@@ -57,6 +57,7 @@ class RoutingInfo(object):
                        "Use a combination of "
                        "get_safe_routing_info_from_pre_vertex, "
                        "get_partitions_outgoing_from_vertex, "
+                       "has_routing_info_from_pre_vertex, "
                        "or get_single_routing_info_from_pre_vertex")
     def get_routing_info_from_pre_vertex(
             self, vertex: AbstractVertex,
@@ -97,6 +98,7 @@ class RoutingInfo(object):
                        "Use a combination of "
                        "get_safe_first_key_from_pre_vertex, "
                        "get_partitions_outgoing_from_vertex, "
+                       "has_routing_info_from_pre_vertex, "
                        "or get_single_first_key_from_pre_vertex")
     def get_first_key_from_pre_vertex(
             self, vertex: AbstractVertex, partition_id: str) -> Optional[int]:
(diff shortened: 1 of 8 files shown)


pacman/model/routing_info/routing_info.py
```diff
@@ -41,46 +43,182 @@ def add_routing_info(self, info: VertexRoutingInfo):
...
+        :param str partition_id:
+            The ID of the partition for which to get the routing information
+        :rtype: VertexRoutingInfo or None
+        """
+        # TODO: Replace (currently temporarily broken to make sure we don't
+        # call it)
+        raise NotImplementedError("Deprecated - shouldn't be used")
+        # return self._info[vertex].get(partition_id)
+
+    def get_safe_routing_info_from_pre_vertex(
```
Christian-B, line 78: or just name it get_info_by_pre

pacman/model/routing_info/routing_info.py
```diff
@@ -41,46 +43,182 @@ def add_routing_info(self, info: VertexRoutingInfo):
...
+        """
+        if vertex not in self._info:
+            return
+        info = self._info[vertex]
+        for partition_id in info:
+            if partition_id not in allowed_partition_ids:
+                raise KeyError(
+                    f"Vertex {vertex} has unknown partition ID {partition_id}")
+
+    def get_single_routing_info_from_pre_vertex(
```
Christian-B, line 183: get__single_info_by_pre

pacman/model/routing_info/routing_info.py
```diff
@@ -41,46 +43,182 @@ def add_routing_info(self, info: VertexRoutingInfo):
...
             The ID of the partition for which to get the routing information
         :return: The routing key of the partition
         :rtype: int
+        :raise KeyError:
+            If the vertex/partition_id combination is not in the routing
+            information
+        """
+        return self._info[vertex][partition_id].key
+
+    def get_partitions_outgoing_from_vertex(
```
Christian-B, line 141: get_partition_from

pacman/model/routing_info/routing_info.py
```diff
@@ -41,46 +43,182 @@ def add_routing_info(self, info: VertexRoutingInfo):
...
+            self, vertex: AbstractVertex) -> Iterable[str]:
+        """
+        Get the outgoing partitions from a vertex.
+
+        :param AbstractVertex vertex: The vertex to search for
+        """
+        return self._info[vertex].keys()
+
+    def has_routing_info_from_pre_vertex(
```
Christian-B, line 150: has_info_from

pacman/model/routing_info/routing_info.py
```diff
@@ -41,46 +43,182 @@ def add_routing_info(self, info: VertexRoutingInfo):
...
+        :param str partition_id:
+            The ID of the partition for which to get the routing information
+        :rtype: bool
+        """
+        if vertex not in self._info:
+            return False
+        info = self._info[vertex]
+        return partition_id in info
+
+    def test_pre_vertex_partition_ids(
```
Christian-B, line 165: check_from

pacman/model/routing_info/routing_info.py
```diff
@@ -41,46 +43,182 @@ def add_routing_info(self, info: VertexRoutingInfo):
...
+        """
+        if vertex not in self._info:
+            return None
+        info = self._info[vertex]
+        if len(info) != 1:
+            raise KeyError(
+                f"Vertex {vertex} has more than one outgoing partition")
+        return next(iter(info.values()))
+
+    def get_single_first_key_from_pre_vertex(
```
Christian-B, line 201: get_single_key_from

rowleya:
Renames done, tests in progress

pacman/model/routing_info/routing_info.py
```diff
@@ -41,46 +43,175 @@ def add_routing_info(self, info: VertexRoutingInfo):
...
+        :param str partition_id:
+            The ID of the partition for which to get the routing information
+        :rtype: bool
+        """
+        if vertex not in self._info:
+            return False
+        info = self._info[vertex]
+        return partition_id in info
+
+    def check_info_from(
```
Christian-B, line 158: While this method is fine I would rather see this functionality moved to add_routing_info

If the  info.vertex knows the allowed_partition_ids before any edges or Projections are added. Check that in add_routing_info which is called when a Projection/ edge is added so offers fail fast!

rowleya: From a user perspective, this will likely not make much of a difference, since I would assume this happens only during a run call.  It will be earlier in the process, but not at time of edge creation, so not much better.  Adding a check at the time of edge creation makes more sense if this is going to be done, but is also perhaps for another PR!

rowleya: (That is, add_routing_info still doesn't happen during Projection / edge creation, only in run)

rowleya: Note also that in reality a *check* should be done at the add time, since a sequence of adding projections may change the outcome in that case e.g. for one vertex, the first partition id "checked" can be the one it will use for its outgoing key, and then any additional ones can be either failures or accepted depending on if the vertex can send with more than one key or not.  It might then accept the names of only certain partitions if it needs to know which partitions are for which reason as well.  A receiver vertex could then also "check" that the given partition id is OK, which could then check if it has already got a key from that vertex, or look into which projection is using the key or ... lots of other options.  This could be quite tricky to implement in general, so probably an implementation per-vertex might be needed, and so definitely doing this in another PR would be recommended!

The idea of the use of the calls being changed is that it is not as complicated for existing vertices.  Yes, it means you don't know until later if you have gone wrong, but given that this is unlikely to be much of a problem most of the time, I don't think that is a big issue!

Christian-B (changes requested):
commented out stuff needs putting back in.

The one more full run inculduing jenkins

The ok to merge

pacman/model/routing_info/routing_info.py
Christian-B: was just an idea not critical.   Can be looked at again when the multi partition stuff comes in

PR #584, opened 2024-10-22
assert not None

Christian-B:
required because of https://github.com/SpiNNakerManchester/SpiNNMachine/pull/273

### DIFF
--- a/pacman/model/tags/tags.py
+++ b/pacman/model/tags/tags.py
@@ -74,8 +74,10 @@ class Tags(object):
             raise PacmanInvalidParameterException(
                 "ip_tag", str(ip_tag), "Only add IP tags with this method.")
         existing_tag = None
-        if (ip_tag.board_address, ip_tag.tag) in self._ip_tags:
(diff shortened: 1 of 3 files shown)


pacman/model/tags/tags.py
```diff
@@ -74,8 +74,10 @@ def add_ip_tag(self, ip_tag: IPTag, vertex: MachineVertex):
             raise PacmanInvalidParameterException(
                 "ip_tag", str(ip_tag), "Only add IP tags with this method.")
         existing_tag = None
-        if (ip_tag.board_address, ip_tag.tag) in self._ip_tags:
-            existing_tag = self._ip_tags[ip_tag.board_address, ip_tag.tag]
+        board_address = ip_tag.board_address
+        assert board_address is not None
```
rowleya, line 78: This seems unnecessary, or else should be handled; unless we know that the board_address is resolved elsewhere?

Christian-B: fixed

PR #587, opened 2024-12-03
Fix routing again

rowleya:
Another routing fix


Christian-B (approved):
ok to merge with our without a unittest

PR #609, opened 2025-07-29
Placer order

Christian-B:
Work to allow the normal placer to happen before the monitors are added.

ApplicationPlacer changes
1. removed _cap_sdram was unneeded and weird as it double counted monitor
2. No longer returns the Placements object just uses it 
3. _space_on_chip adds monitor  costs to check the chip has space

tests changed as ApplicationPlacer no longer returns


Christian-B:
It appears that ApplicationPlacer._place_error compares the previous placements with the current placements.
However for speed master no longer clones the placements so this does not amke sense.

Options are.
1. Leave the code as in with a log that the code should eb changed if needed.
2. Clone the previous placements 
  - always (previous will now often be empty_
3. Clone the previous on a cfg setting for example Mode = debug
4. Strip that compare code out

PR #622, opened 2026-02-27
rework routing allocator and provide more data

Christian-B:
ZonedRoutingInfoAllocator updated

-flexible option removed as the global one handles as many cases better
-This is now very similar to the previous global

- zones values given better names and a few extra added to make code clearer
- zone values added to RoutingInfo
- overlap app vetex, patition pairs added to RoutingInfo
   - ap will be only some fixed
   - mx will be ALL fixed and some vertices with many atoms

Fixed Routes are do at the start avoiding the need to cash data
- cleaned up to be clearer that for each fapp vertex
  - The App must specify fix if any machine does
  - For App with more than 1 fixed each must specify fixed

The Application/Partition zone is now always at the top and as small as possible
- just the bits needed for 1 key per AppVetrex/ parition Id pair
- This zone will be enlarge if there is a lot of fixed zone overlaps
- extra bits will be in the machine zone

Clearer that (except fixed) all vertices will respect the AP zone

Test adjusted to show zone

Removed test where when both flexible and zone where tested with the same graph

### DIFF
--- a/pacman/model/routing_info/routing_info.py
+++ b/pacman/model/routing_info/routing_info.py
@@ -314,7 +314,7 @@ class RoutingInfo(object):
         """
         Checks if there are any atoms into machine zone overlaps
 
-        Trie if there are any fixed key mask vertexes.
+        True if there are any fixed key mask vertexes.
 
         :return: True if and only if there is at least one overlap
         """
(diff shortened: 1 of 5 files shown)


Christian-B:
replaced by https://github.com/SpiNNakerManchester/PACMAN/pull/625

PR #625, opened 2026-03-23
Right fixed keys

Christian-B:
builds on / replaces https://github.com/SpiNNakerManchester/PACMAN/pull/623 and https://github.com/SpiNNakerManchester/PACMAN/pull/622

Changes the routing info allocator.

All infos will work with the same Application (partition Id) shit!

All infos will be machine shiftable.
 - may not be the same for all infos
 
The Allocator now only uses the Global approach in master
- all are the same zones except when there is a mix of retinas and high machine index 

Fixed keys must work with the rest or an Exception is raised.

- The mask MUST be shiftable. (all one bits higher than zero bits)
   - Therefor they must be app zone. machine zone, atom zone top to bottom

- If there is a single fixed application mask that can be used it is
- If there is also a single fixed machine mask that can be used it is used too

- Fixed masks MUST work with a GLOBAL atom shift.
   - There is a one to one relationship zones may be split in a place not suggested by the fixed.
      - This includes ones like the CommandSender with 0xFFFFFFFF masks
   - If The none zero bit are high for app and low for machine index the split can be anywhere in the middle
   
There may be multiple fixed App Vertex/ partition pairs with the same app key after apply the global shift
- Because they actually have the same fixed app key
- Because the global shift makes them the same

 support methods  both on each Info and the RoutingInfo collection Object
- has_fixed_keys 
- has_global_masks
- has_global_app_masks
- has_app_keys_overlap

---
Changes included

Different Classes for Global, Specific and Fixed RoutingInfo

Added method VertexRoutingInfo.app_vertex  

RouutingInfo
 - removed deprecated values
 - Added Allocator values used.
 
Utils
 get_key_ranges no longer needed due to fixed must be shiftable

tests
- Use SpecificMachineVertexRoutingInfo as MachineVertexRoutingInfo is now abstract
- Keys and Masks changed to fit new rules
- Tests for each RoutingInfo class
- Adapted Allocator tests
-


Christian-B:
Fixes https://github.com/SpiNNakerManchester/PACMAN/issues/626 by replacing n_bits_atoms with machine_shift which fails on some fixed keys

pacman/utilities/utility_calls.py
```diff
@@ -183,3 +147,93 @@ def is_power_of_2(v: int) -> bool:
...
+
+    If key is zero will return 32
+
+    :returns: index with the most significant bit being 0
+    """
+    bits = expand_to_bit_array(key)
+    for i in range(BITS_IN_KEY):
+        if bits[i] == 1:
+            return i
+    return 32
```
rowleya: Isn't 32 == BITS_IN_KEY (i.e. can you use that constant)?

pacman/utilities/utility_calls.py
```diff
@@ -183,3 +147,93 @@ def is_power_of_2(v: int) -> bool:
...
+
+    This requires a mask where all the 1 bits are before all the zero bits
+
+    :param mask:
+    :return: The shift
+    :raises PacmanValueError: If the mask does not support a clean shift
+    """
+    bits = expand_to_bit_array(mask)
+    found_shift = False
+    shift = -1000
```
rowleya, line 164: In python you don't need to pre-declare the variable that you will set later, or otherwise not use I think.

pacman/utilities/utility_calls.py
```diff
@@ -183,3 +147,93 @@ def is_power_of_2(v: int) -> bool:
...
+    for i in range(32):
```
rowleya: 32 == BITS_IN_KEY

pacman/utilities/utility_calls.py
```diff
@@ -183,3 +147,93 @@ def is_power_of_2(v: int) -> bool:
...
+def can_shift(mask: int) -> bool:
+    """
+    Checks if the mask can generate a clean shift.
+
+    :param mask:
+    :return: True if cal_shift will work or False if it will error.
+    """
+    bits = expand_to_bit_array(mask)
+    found_zeros = False
+    for i in range(32):
```
rowleya: 32 == BITS_IN_KEY

pacman/operations/routing_info_allocator_algorithms/zoned_routing_info_allocator.py
```diff
@@ -86,333 +95,390 @@ class ZonedRoutingInfoAllocator(object):
...
+                raise PacmanRouteInfoAllocationException(
+                    f"For partition {part_id} {pre} has fixed key "
+                    f"{app_key_and_mask} "
+                    f"while outgoing {m_vertex} has no fixed key")
+            if (key_and_mask.key & app_key_and_mask.mask !=
+                    app_key_and_mask.key):
+                raise PacmanRouteInfoAllocationException(
+                    f"For partition {part_id} {pre} has fixed key "
+                    f"{app_key_and_mask} "
+                    f"while outgoing {m_vertex} has {key_and_mask}"
```
rowleya: Missing space

rowleya (changes requested):
Minor changes requested, but looks OK overall

pacman/utilities/utility_calls.py
Christian-B: Pythonically you are right.

But pylint doed not like it

E0601: Using variable 'shift' before assignment (used-before-assignment)

PR #627, opened 2026-03-30
Validate routes uncompressed

### LINKED ISSUE #442: valid_routes_checker use or loose
pacman/operations/multi_cast_router_check_functionality/valid_routes_checker.py

No use case found.

Christian-B:
fixes https://github.com/SpiNNakerManchester/PACMAN/issues/442

range_masks should not include the mass 0xFFFFFFFF
These are used by the CommandSender and so are a special case

AbstractVirtual will not have routes.

Tested with https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1348 as validate_routes_uncompressed = True.

Now debug so it only runs as needed

### DIFF
--- a/pacman/operations/multi_cast_router_check_functionality/valid_routes_checker.py
+++ b/pacman/operations/multi_cast_router_check_functionality/valid_routes_checker.py
@@ -23,17 +23,19 @@ from spinn_utilities.log import FormatAdapter
 from spinn_machine import Chip, MulticastRoutingEntry
 from pacman.data import PacmanDataView
 from pacman.exceptions import (
-    PacmanConfigurationException, PacmanRoutingException)
+    PacmanConfigurationException, PacmanNotPlacedError, PacmanRoutingException)
 from pacman.model.graphs.application import ApplicationVertex
 from pacman.model.graphs import AbstractVirtual
 from pacman.utilities.constants import FULL_MASK
 from pacman.utilities.algorithm_utilities.routing_algorithm_utilities import (
     get_app_partitions)
-from pacman.model.graphs.machine import MachineVertex
+from pacman.model.graphs.machine import (
+    MachineFPGAVertex, MachineSpiNNakerLinkVertex, MachineVertex)
 from pacman.model.placements import Placement
 from pacman.model.routing_info import BaseKeyAndMask
 from pacman.model.routing_tables import (
     AbstractMulticastRoutingTable, MulticastRoutingTables)
+
 
 logger = FormatAdapter(logging.getLogger(__name__))
 range_masks = {FULL_MASK - ((2 ** i) - 1) for i in range(33)}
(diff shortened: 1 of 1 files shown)


Christian-B:
currently fails with spynnaker_integration_tests/test_split_various/test_split_delays.py

rowleya:
This is down to how the splitter works and because it isn't caching the result.  When ```get_source_specific_in_coming_vertices``` is called in this case, the splitter has an internal variable that it uses to rotate between the synapse cores so that the incoming traffic is roughly evenly split over the cores.  This happens [here](https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/extra_algorithms/splitter_components/splitter_population_vertex_neurons_synapses.py#L543-L545) (and note that in a [different branch](https://github.com/SpiNNakerManchester/sPyNNaker/blob/spike_filter/spynnaker/pyNN/extra_algorithms/splitter_components/splitter_population_vertex_neurons_synapses.py#L600-L601) this is happening slightly differently also).  The fix would be to add some sort of cache so that the first index is taken from the cache if the same source vertex is presented, otherwise use the global value.

pacman/operations/multi_cast_router_check_functionality/valid_routes_checker.py
```diff
@@ -35,8 +35,9 @@
 from pacman.model.routing_tables import (
     AbstractMulticastRoutingTable, MulticastRoutingTables)
 
+
 logger = FormatAdapter(logging.getLogger(__name__))
-range_masks = {FULL_MASK - ((2 ** i) - 1) for i in range(33)}
+range_masks = {FULL_MASK - ((2 ** i) - 1) for i in range(1, 33)}
```
rowleya, line 40: Not sure this is the right answer... it should be possible for routes to use that mask and have this still work...

pacman/operations/multi_cast_router_check_functionality/valid_routes_checker.py
```diff
@@ -88,6 +89,8 @@ def validate_routes(routing_tables: MulticastRoutingTables) -> None:
                     source, partition.identifier)
 
             for tgt, srcs in target_vertices:
+                if isinstance(tgt, AbstractVirtual):
+                    continue
```
rowleya, line 93: It would still be useful to validate that virtual routes go where they should.  The route should go to the chip that the virtual vertex is connected to (or come from that chip) and then have one additional link that takes it to the virtual vertex.  In that case the destination can be the real chip that the vertex is connected to.  An additional check should then be done to determine if the final route is also there to make sure that the traffic goes to where it should.

pacman/operations/multi_cast_router_check_functionality/valid_routes_checker.py
rowleya, line 122: This is then also wrong (this is existing code), since it should be possible to use this as a source, just getting the real placement of the real chip the virtual vertex is connected to as the starting point.  In this case, there is no additional check needed - the packets should just appear on this chip from "nowhere".

rowleya (changes requested):

pacman/operations/multi_cast_router_check_functionality/valid_routes_checker.py
Christian-B: range_masks should not include the mass 0xFFFFFFFF
These are used by the CommandSender and so are a special case

rowleya: I think they are needed, but probably need the AbstractVirtual targets enabled also.  Command sender routes should work after all!

PR #629, opened 2026-05-21
no routes to AbstractVirtual

Christian-B:


rowleya (approved):
Happy to approve as blocker fix, but noting this is also part of #627 and requested to change there!

PR #633, opened 2026-06-02
simplify by removing @pytest.fixture

Christian-B:
The pytest.fixture is not needed as it there was only 1 option.


rowleya (approved):
Fair enough - I guess we may have tried different parameters previously or something?

PR #635, opened 2026-06-08
Spin gen enums

Christian-B:
part of https://github.com/SpiNNakerManchester/SpiNNMachine/pull/316

also replaces hard coded version numbers and where possible tests both spin1 and spin2


Christian-B:
Agreed with @rowleya to instead make all these tests parameterized

Christian-B:
IN most of these case ONLY a version 5 boards works as Spinnaker link or FPGA's used

https://github.com/SpiNNakerManchester/PACMAN/pull/634 overrides the ones that can be multiple boards

rowleya (approved):

